PR #194, opened 2023-06-29
Fix empty preimage

roeierez:
We get empty `preimage` for payments retrieved via `list_payments` endpoint.
By observing the node client `list_pays` endpoint it seems that the value there is correct but renamed differently causing the deserialization result in empty string.
I don't have a way to test it currently but pretty sure this will fix it.


cdecker:
Good catch, thanks :-)

PR #204, opened 2023-07-10
Reconnect both when a signer attaches and when the RPC becomes available

cdecker:
We were having issues with nodes not reconnecting some times. We traced the issue to the signer connecting very early on, when the RPC is not available yet, causing the `connect()` calls yelling into the void.

By refactoring the reconnection logic and calling it both from the RPC watcher and the signer connecting should address that.

### DIFF
--- a/libs/gl-plugin/src/node/mod.rs
+++ b/libs/gl-plugin/src/node/mod.rs
@@ -34,6 +34,11 @@ static LIMITER: OnceCell<RateLimiter<NotKeyed, InMemoryState, MonotonicClock>> =
 
 lazy_static! {
     static ref HSM_ID_COUNT: AtomicUsize = AtomicUsize::new(0);
+
+    /// The number of signers that are currently connected (best guess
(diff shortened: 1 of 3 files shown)


libs/gl-plugin/src/node/mod.rs
```diff
@@ -936,54 +867,125 @@ use nix::unistd::getppid;
...
+	    let res = rpc.call_typed(r.clone()).await;
+	    trace!("Connect returned: {:?} -> {:?}", &r.id, res);
+
+	    match res {
+		Ok(r) => info!("Connection to {} established: {:?}", &r.id, r),
+		Err(e) => warn!("Could not connect to {}: {:?}", &r.id, e),
+	    }
+	}
+
+	todo!()
```
nepet: I suspect this `todo` was forgotten to remove

nepet (approved):
Seems good to me!
Ack 17c19804f5fc0727b4ea7ea4ec316834beaf1b3a

PR #206, opened 2023-07-12
Lightning Service Provider Specification

ErikDeSmedt:
Draft on how to implement the LSPS-specification in greenlight.

Todo:
- Write integration tests
- Detect Lightning Service Providers by lookint at feature bit 729


cdecker (commented):
Very nice JSON-RPC implementation :+1: Some of the concepts are very cool, and I might backport some of them into `cln-rpc` which also has a tiny JSON-RPC implementation.

cdecker:
I think if you have a small test that sends and receives messages from one node to the other we can already merge this, so we can pipeline changes in and avoid large PRs :-)

PR #214, opened 2023-07-18
Add runes

nepet:
This PR is the first in a row of PRs to enable the Device-Signer-Pairing process. It is intended to enable the use of [`runes`](https://github.com/nepet/futhark) as authorization tokens on the clients signer.

- [x] Derive the runes secret base from the signer seed
- [x] Add functionality to create a new rune based on the main rune of the signer
- [x] Pass rune through http headers and context from the client libs to the signer 
- [x] Pass rune through the py and js bindings
- [x] Add the public key that is used for the mTLS authentication to the rune 
- [x] Check that a rune is valid on signing requests

Apologies for the maintenance and refactoring commits that slipped in.


libs/gl-plugin/src/node/mod.rs
```diff
@@ -1127,10 +1129,11 @@ where
                 }
 
                 trace!(
-                    "Got a request for {} with pubkey={}, sig={} and body size={:?}",
+                    "Got a request for {} with pubkey={}, sig={}, rune={} and body size={:?}",
```
cdecker, line 1132: Notice that for this to be safe (i.e., node can't steal the rune and use it for its own commands) we need to either commit to the request in the Rune (append the command as a constraint at the end of the rune) or have the run commit to the pubkey used to sign the request. The latter is more flexible as it doesn't require keeping the raw rune around on the client, which might leak accidentally, or require the serialized payload at rune-verification time.

cdecker: I think you'll likely have considered that in the followup commit by the way, I just wanted to make sure that I check it myself :-)

libs/gl-client/src/node/service.rs
```diff
@@ -136,6 +142,11 @@ impl Service<Request<BoxBody>> for AuthService {
                 .headers
                 .insert("glts", engine.encode(ts).parse().unwrap());
 
+            // Runes already come base64 URL_SAFE encoded.
+            parts
+                .headers
+                .insert("glrune", rune.parse().expect("Could not parse rune"));
+
```
cdecker, line 149: Just passing the rune through here, so the Rune must be committing to the pubkey, and the signer needs to verify that the pubkey for the signature and the commitment in the rune match.

libs/proto/scheduler.proto
```diff
@@ -222,6 +222,11 @@ message RegistrationResponse {
 	// The private key that was used to create the certificate with. This key
 	// is used to sign the requests to the node.
 	string device_key = 2;
+
+	// A master rune that is returned if the device that is registered has its
+	// own signer. The signer is necessary as the response is intercepted on the
+	// client side and appends the rune to the registratrion response.
+	string rune = 3;
```
cdecker, line 229: Ah, this is where we have the divergence we discussed between scheduler <-> client comms and gl-client <-> caller. The rune isn't returned by the scheduler, it is being backfilled by the signer when the result comes back in the same way the CSR -> CRT backfilling works already :+1:

cdecker (commented):
Very good, just some minor comments regarding the security of how we handle Runes and that we still need to constrain it at least to commit to the pubkey, and the signer needs to verify that constraint as well.

libs/proto/scheduler.proto
nepet: Exactly :)

libs/gl-plugin/src/node/mod.rs
nepet: Always good to double check on critical parts :+1: 

We already build the rune with the public key in the `id` field of the rune. This binds the rune to the public key. The rune is based on the `authcode` of the signing device, that itself is based on a derived key (hkd) from the signer. This should commit us to the public key as well as attest the public key.

But as you said, we could of course even go one step further and use _one-time-runes_ (otrs ^^) and commit the rune to the signed request by adding the signature to the rune.

libs/gl-client/src/node/service.rs
nepet: Yes, validating the rune against the public key -> signature -> request should be the last part that is missing for this pull request.

libs/gl-client/src/signer/mod.rs
```diff
@@ -260,6 +270,9 @@ impl Signer {
             let signer_state = req.signer_state.clone();
             trace!("Received request {}", hex_req);
 
+            debug!("Here!");
+            debug!("{:?}", req.requests);
+
```
cdecker: Debug logging that was forgotten? :p

libs/gl-client/src/signer/mod.rs
```diff
@@ -592,6 +605,29 @@ impl Signer {
     pub fn version(&self) -> &'static str {
         VERSION
     }
+
+    /// Create a base64 encoded rune from the master rune of this signer.
+    pub fn create_rune(
+        &self,
+        unique_id: Option<&str>,
+        restrictions: Option<&str>,
```
cdecker, line 661: I wonder if passing in the correct types rather than their stringified version wouldn't be better. This way we're likely going to create restrictions, stringifying them, then parsing them again. Also by passing in the correct types we make this pretty much infallible.

libs/gl-client/src/signer/mod.rs
```diff
@@ -225,6 +225,7 @@ impl Signer {
                 if r.timestamp != 0 {
                     data.put_u64(r.timestamp);
                 }
+                data.put(&r.rune[..]);
```
cdecker: This is a breaking change: changing the way we verify signatures means that we can't mix pre-change clients and post-change signers and vice-versa. This is not an issue in this case because we have a rather tight coupling still, but better keep this in mind for when we have keyless clients as they won't upgrade in lockstep.

libs/gl-client-py/glclient/glclient.pyi
```diff
@@ -15,6 +15,7 @@ class Signer:
     def version(self) -> str: ...
     def is_running(self) -> bool: ...
     def shutdown(self) -> None: ...
+    def create_rune(self) -> str: ...
```
cdecker, line 18: Seems this is missing, at least in this commit.

libs/gl-client-py/tests/test_plugin.py
```diff
@@ -30,4 +30,10 @@ def test_max_message_size(clients):
...
     with pytest.raises(ValueError):
-        n1.datastore("some-key", hex=bytes.fromhex(secrets.token_hex(size)))
\ No newline at end of file
+        n1.datastore("some-key", hex=bytes.fromhex(secrets.token_hex(size)))
+
+def test_node_info(clients):
+    c1 = clients.new()
+    c1.register()
+    n1 = c1.node()
+    print(n1.get_info())
```
cdecker, line 39: Aren't there already several tests that do this and more? While I share the desire to have a test have a single responsibility that is not possible for slow integration tests.

libs/gl-client-py/glclient/__init__.py
```diff
@@ -55,11 +55,12 @@ def is_running(self) -> bool:
 
 class Scheduler(object):
 
-    def __init__(self, node_id: bytes, network: str, tls: TlsConfig):
+    def __init__(self, node_id: bytes, network: str, tls: TlsConfig, rune: str=""):
```
cdecker, line 58: Setting a default does not sounds like a good idea, as it suggests that users don't need to set this, and they'll run into issues once we start enforcing these constraints.

We should rather start consolidating all of these arguments into a single blob that can be upgraded from version to version. A protobuf struct could serve for this: we hand out a blob, and expect it to be handed back when loading, so the callers just need to persist however they want and not care about the details.

libs/gl-testing/gltesting/clients.py
```diff
@@ -74,10 +74,24 @@ def tls(self) -> glclient.TlsConfig:
...
+    
+    def rune(self) -> str:
+        """Load the devices rune
+        
+        Returns the devices rune if it is available or an empty string
+        if the rune can not be found.
+        """
+        self.log.info("Trying to find a rune")
+        runepath = self.directory / "device.rune"
+        rune = ""
```
cdecker, line 86: Empty runes must fail. Especially when we have control over the entire environment which is the case for us in `gl-testing`. If we do not enforce correctness in what we test, then how can we expect our users to?

libs/gl-client/Cargo.toml
```diff
@@ -35,7 +35,7 @@ serde_json = "^1.0"
 thiserror = "1"
 cln-grpc = { git = "https://github.com/ElementsProject/lightning.git", branch = "master"}
 chacha20poly1305 = { version = "0.10.1", optional = true }
-futhark = { git = "https://github.com/nepet/futhark.git", branch = "main"}
+futhark = { git = "https://github.com/nepet/futhark.git", commit = "611bcab"}
```
cdecker, line 38: Is this on purpose? Pointing at a branch sounds safer since that can move and sync up with the branch, whereas pointing at a commit may cause confusion when developing on the remote repo, but changes don't land in the client.

libs/gl-client/src/signer/mod.rs
```diff
@@ -225,15 +230,55 @@ impl Signer {
...
                     .map(|_| r)
-                    .map_err(|e| anyhow!("signature verification failed: {}", e))
+                    .map_err(|e| anyhow!("rune verification failed: {}", e))
             })
             .collect()
     }
 
+    /// Verifies that the public key of the request and the signers rune version
+    /// match the corresponding restrictions of the rune.
+    fn verify_rune(&self, request: crate::pb::PendingRequest) -> Result<(), anyhow::Error> {
```
cdecker, line 248: Notice that you probably want the decoded PB payload, otherwise you will have a hard time checking the rune against the undecoded commands :-)

libs/gl-client/src/signer/mod.rs
```diff
@@ -225,15 +230,55 @@ impl Signer {
...
+        let rune64 = general_purpose::URL_SAFE.encode(request.rune);
+        let rune = Rune::from_base64(&rune64)?;
+
+        // A valid gl-rune must contain a pubkey field as this  is bound to the
+        // signer. Against the rules of runes we do not accept a rune that has
+        // no restriction on a public key.
+        if !rune.to_string().contains("pubkey=") {
+            return Err(anyhow!("rune is missing pubkey field"));
+        }
```
cdecker, line 257: Is this not redundant with the check directly below? Doing a string search on the unparsed restrictions sounds like it can easily be tricked by the client just appending something that contains the substring you are searching for.

libs/gl-client/src/signer/mod.rs
```diff
@@ -636,6 +678,25 @@ impl Signer {
...
 }
 
+/// Used to extract the unique id from a rune. This is used to build
+/// a version check in the case that a rune has set a unique id and a
+/// version.
+/// Todo: Piggiback into runes -> A rune shall return the id.
+fn extract_device_id(rune64: &str) -> Option<String> {
+    Rune::from_base64(rune64)
+        .ok()
+        .map(|r| r.to_string())
```
cdecker, line 688: We parse the base64 string to then stringify it again and doing string operations on the entire string? Would iterating over the restrictions and extracting the key-value pairs from there not be safer? If we do string ops on the entire string we might get tricked by a clever client adding the desired substrings to the end of the rune.

cdecker (commented):
Very good PR :+1: I went through with a more attacker focused approach and I found a couple of places where an attacker client could inject substrings that can trick the signer since it is using string operations to extract and verify some pieces of the rune.

Not sure if they are really exploitable, but since we have the rune as parsed structs that could be used to extract and verify much more easily, we could/should probably use those.

libs/gl-client/src/signer/mod.rs
nepet: To be fair, this check is pretty useless if an attacker gets access to an *unrestricted* rune. This check should just ensure that we do not use unrestricted runes by accident (in development, while testing etc):
The thing with runes being a set of restrictions is: If the restriction is not present, we would not check on it, so a rune that does not have a `pubkey` field present would pass the second check as we would simply skip the check. So the first check ensures that the second check will always be performed, while the second check ensures that the public key of the request matches the restriction of the rune, performing the check on the rune itself and not the unparsed restriction.

>Doing a string search on the unparsed restrictions sounds like it can easily be tricked by the client just appending something that contains the substring you are searching for.

That is correct and the reason why the length-extension-attack-feature of runes only allows them to be a set of restrictions, rather than allowances. But as said above, this is only a problem if an attacker has access to an unrestricted rune and this is the reason why we must be especially careful that we do not distribute runes that are not restricted to the `pubkey` of the device.

libs/gl-client/src/signer/mod.rs
nepet: In our case, this is only a problem if we ever hand out a unrestricted rune. One can only ever append to a rune and keep the rune authorized (runes are a set of restrictions, not allowances). Internally the rune is parsed the same way reading the restrictions from the string in the first place. The base64 string is just an encoding of the rune string so we could also skip creating a rune and just decode the base64 string the rune string directly.

But yes I am on your page, there are some features I want to add to the rune library, e.g. having a method that returns the restrictions as well as a method that returns the id and/or the version that is set for a rune.

libs/gl-client/src/signer/mod.rs
nepet: :+1:

libs/gl-client/Cargo.toml
nepet: Good catch! This was to avoid caching issues, I will change it back to master :+1:

libs/gl-client/src/signer/mod.rs
nepet: My bad :facepalm: :sweat_smile:

libs/gl-client-py/glclient/glclient.pyi
nepet: Yes thanks for noticing! This is dead code from early experiments as we do not hand out runes so far.

libs/gl-client/src/signer/mod.rs
nepet: Oh I forgot that I removed the `create_runes` function from the clients ^^. Yes I can switch back using correct types. Still we might want to think about the enums as signer-less devices will need to ask for restrictions during the pairing and this is something we will likely initiate from the bindings-layer.

libs/gl-client-py/glclient/__init__.py
nepet: Yes I already had this in mind as a next step in a separate PR. I could also add it in this PR if you think it is more convenient to have these changes tied together so that people only need to `recover` once.
In this case I can also add these changes to the `cli` tool with a descriptive info/error message.

cdecker:
Adding this to the `main-0.2` branch as target, so we can collect breaking changes in one go.

libs/gl-testing/gltesting/clients.py
nepet: The empty rune is used on registration and recovery, that is why we set it to an empty string here. We might want to add a check if the nobody identity is used and fail early if not and if the rune file can not be found. What do you think?

cdecker:
ACK https://github.com/Blockstream/greenlight/pull/214/commits/ece5295a504d04daaadc4f14fcd0788098941643

Issue #216, opened 2023-07-19
Drop support for python3.7 in gl-client-py


ErikDeSmedt:
The `glclient` package is currently not functional when used with `python3.7`.
The `SupportsIndex` type was introduced python 3.8 and we use it [here](https://github.com/Blockstream/greenlight/blob/99424f36073433803bfafe33c7920e3568c3b9a7/libs/gl-client-py/glclient/__init__.py#L10C53-L10C66).

I don't think we should support python versions that are end-of-life [end-of-life](https://devguide.python.org/versions/).
Therefore, I suggest we move to 3.9

cdecker:
Addressed by #218

PR #217, opened 2023-07-19
Drop support for python versions that are EOL

ErikDeSmedt:
The libraries are not working for python 3.7 because we are using the `SupportsIndex` type.

Version 3.7 and 3.8 are also end-of-life.

Resolves
[216](https://github.com/Blockstream/greenlight/issues/216)

### DIFF
--- a/libs/gl-client-py/pyproject.toml
+++ b/libs/gl-client-py/pyproject.toml
@@ -22,7 +22,7 @@ packages = [
 black = "^23.1.0"
 
 [tool.poetry.dependencies]
-python = ">=3.7,<4"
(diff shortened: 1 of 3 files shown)


cdecker:
I'd rather keep 3.7 and avoid using the `SupportsIndex` type if that's the only breaking change. We still have plenty of users that are not on Py 3.9, and it'd be a rather blunt limitation, if we were to raise the minimum supported version just for type annotations.

ErikDeSmedt:
Okay, I'll replace this by [MR 218](https://github.com/Blockstream/greenlight/pull/218)

PR #222, opened 2023-07-26
signer: expose sign_message with recovery id

JssDWt:
Expose the `sign_message` function on the signer, in order to allow clients to sign messages themselves. `sign_message` now also returns the recovery id, in order to provide full capabilities of encoding a lightning signed message to external people.


JssDWt:
@cdecker Is this something you're willing to merge. We've got a [PR here](https://github.com/breez/breez-sdk/pull/319) to expose sign_message and verify_message in the Breez SDK. It would bee great to be able to include this functionality. (I can't ask for review I think)

cdecker:
Apologies for the delay, I will also publish new bindings for the languages this week, since they are lagging a bit now :-)

PR #225, opened 2023-07-31
[Draft] Added code that implements the lnurl lud06 spec.

Randy808:


Randy808:
Closing because the commit in this PR is not updating when I push to my branch

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    println!("Domain: {}", Url::parse(url).unwrap().host().unwrap());
+    println!("Description: {}", description);
+    println!(
+        "Accepted range (in millisatoshis): {} - {}",
+        lnurl_pay_request.min_sendable, lnurl_pay_request.max_sendable
+    );
+
+    let callback_url: &str =
+        &format!("{}?amount={}", lnurl_pay_request.callback.clone(), amount).to_string();
```
tkijewski, line 59: heads up here, this will break if url has existing query params. common issue in lnurl implementations

PR #226, opened 2023-07-31
client: Added code that implements the lnurl lud06 spec.

Randy808:


libs/gl-client/Cargo.toml
```diff
@@ -10,21 +10,28 @@ export = ["chacha20poly1305", "secp256k1"]
...
+bech32 = "0.9.1"
 bytes = "1.2.1"
 hex = "*"
 http = "0.2"
 http-body = "^0.4"
+lightning-invoice = "0.23.0"
 log = "^0.4"
 pin-project = "*"
 prost = "0.11"
+reqwest = {version="0.11.18", features=["blocking", "json"]}
```
cdecker, line 24: Everything we do is `async`, why are you using `blocking` here? It will block the async reactor thread arbitrarily while the request is in flight.

libs/gl-client/src/lnurl/mod.rs
```diff
@@ -0,0 +1,15 @@
+mod pay;
+mod tor_http_client;
+
+use anyhow::{Result};
+use pay::resolve_to_invoice;
+use pay::models::{LnUrlHttpClientImpl};
+
+pub struct LNURL;
```
cdecker, line 8: Probably will need to have access to a `ClnClient` so it can trigger operations on the `Node`.

libs/gl-client/src/node/mod.rs
```diff
@@ -1,5 +1,3 @@
-use std::time::Duration;
-
```
cdecker, line 2: Thanks for doing this cleanup. It may be better to split drive-by fixes like this into a separate commit, allowing us to cherry-pick them onto a bug-fix branch, and avoid conflicts if multiple people get annoyed by the lint warning and fix it (I did that in 3 branches so far, and rebase will just deduplicate the cherry-picked commits if we do it correctly).

libs/gl-client/src/lnurl/tor_http_client.rs
```diff
@@ -0,0 +1,48 @@
...
+    tor_client: TorClient<tor_rtcompat::PreferredRuntime>,
+}
+
+impl TorHttpClient {
+    pub async fn new() -> Result<Self> {
+        let mut config_builder = TorClientConfig::builder();
+        let address_filter = config_builder.address_filter();
+        address_filter.allow_onion_addrs(true);
+        address_filter.build()?;
+        config_builder.build()?;
```
cdecker, line 17: This call appears to be a duplicate of the one in the `create_bootstrapped` call, which uses `.unwrap()`. It'd be better to just pull this into the call below instead of doing a pre-flight check like this.

libs/gl-client/src/lnurl/tor_http_client.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        let tor_client = TorClient::create_bootstrapped(config_builder.build().unwrap()).await?;
```
cdecker, line 19:
```suggestion
        let tor_client = TorClient::create_bootstrapped(config_builder.build()?).await?;
```

libs/gl-client/src/lnurl/tor_http_client.rs
```diff
@@ -0,0 +1,48 @@
...
+        Ok(Self { tor_client })
+    }
+
+    pub async fn tor_http_get(&mut self, url: Url) -> Result<()> {
+        info!("Connecting to onion service. This may take a while...");
```
cdecker, line 24:
```suggestion
        debug!("Connecting to onion service. This may take a while...");
```

libs/gl-client/src/lnurl/tor_http_client.rs
```diff
@@ -0,0 +1,48 @@
...
+            .unwrap();
+
+        info!("Connecting to {}...", url.to_string());
+        let http_message = format!(
+            "GET {} HTTP/1.1\r\nHost: {}\r\n\r\n",
+            url.path(),
+            url.domain().unwrap()
+        )
+        .as_bytes()
+        .to_owned();
```
cdecker, line 38: Does `reqwest` not allow setting up a separate `Connector` that is Tor-aware? Doing an open-coded HTTP implementation here is probably out of scope (missing content-type handling, chunking, multipart return, TLS support, ...).

libs/gl-client/src/lnurl/tor_http_client.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct TorHttpClient {
+    tor_client: TorClient<tor_rtcompat::PreferredRuntime>,
+}
+
+impl TorHttpClient {
+    pub async fn new() -> Result<Self> {
+        let mut config_builder = TorClientConfig::builder();
+        let address_filter = config_builder.address_filter();
+        address_filter.allow_onion_addrs(true);
+        address_filter.build()?;
```
cdecker, line 16: The `address_filter` is not used anywhere? And the fact that `address_filter` is retrieved from `config_builder` but not `mut` suggests that any change to it will not be reflected in `config_builder`. Or the API might just be strange :-)

libs/gl-client/src/lnurl/pay/models.rs
```diff
@@ -0,0 +1,88 @@
...
+                e
+            )),
+            Ok(e) => Ok(e),
+        }
+    }
+}
+
+impl LnUrlHttpClient for LnUrlHttpClientImpl {
+    fn get_pay_request_response(&self, lnurl: &str) -> Result<PayRequestResponse> {
+        Ok(self.get::<PayRequestResponse>(lnurl).unwrap())
```
cdecker, line 76:
```suggestion
        Ok(self.get::<PayRequestResponse>(lnurl)?)
```
A lot can go wrong on the public internet, but only a few things are fatal, so allow the caller to recover.

libs/gl-client/src/lnurl/pay/models.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    fn get_pay_request_callback_response(
+        &self,
+        callback_url: &str,
+        amount: u64,
+    ) -> Result<PayRequestCallbackResponse> {
+        Ok(self
+            .get::<PayRequestCallbackResponse>(callback_url)
+            .unwrap())
```
cdecker, line 86:
```suggestion
            .get::<PayRequestCallbackResponse>(callback_url)?)
```

libs/gl-client/src/lnurl/pay/models.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn new() -> LnUrlHttpClientImpl {
+        let tokio_runtime = Runtime::new().expect("Failed to create Tokio runtime");
+        let tor_client = tokio_runtime.block_on(TorHttpClient::new()).unwrap();
+        LnUrlHttpClientImpl {
+            clearnet_client: reqwest::Client::new(),
+            tor_client,
+        }
+    }
+
+    fn get<T: DeserializeOwned + 'static>(&self, url: &str) -> Result<T> {
```
cdecker, line 55: If we make this `async` we do not have to build a nested `Runtime`.

libs/gl-client/src/lnurl/tor_http_client.rs
cdecker: Notice too that `create_bootstrapped` is actually going out and establishing Tor circuits, which can take considerable time. If we then end up not using that circuit all of that work is wasted. It'd likely be a better idea to keep the builder around, and defer bootstrapping to when we know we need the circuit.

libs/gl-client/src/lnurl/pay/models.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub struct LnUrlHttpClientImpl {
+    clearnet_client: reqwest::Client,
+    tor_client: TorHttpClient,
+}
+
+impl LnUrlHttpClientImpl {
+    /// Constructs a new `LnUrlHttpClient`.
+    ///
+    /// # Panics
```
cdecker, line 43: It should not panic :-)

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// Validates the invoice on the pay request's callback response
+fn validate_invoice_from_callback_response(
+    invoice: &Invoice,
+    amount: u64,
+    lnurl_pay_request: PayRequestResponse,
+) -> Result<()> {
+    ensure!(invoice.amount_milli_satoshis().unwrap_or_default() == amount ,
+        format!("Amount found in invoice was not equal to the amount found in the original request\nRequest amount: {}\nInvoice amount:{:?}", amount, invoice.amount_milli_satoshis().unwrap())
```
cdecker, line 125: The `ensure!` macro does support string formatting, as far as I understand, so you don't need to nest a `format!()` call into it.

cdecker (commented):
Wow `arti` is huge :-) Slowly getting concerned about the binary size that we have to ship to various bindings. As promised I went through it with a really fine-toothed comb, so there are a lot of comments, but feel free to ignore them.

The mix of `async` and `blocking` is a bit strange, and some operations, such as initializing Tor circuits, could be deferred a bit so we don't pay the costs even though we don't use it later.

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: 👍

libs/gl-client/src/lnurl/pay/models.rs
Randy808: 👍

libs/gl-client/src/lnurl/pay/models.rs
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    /// This method panics if a http or tor client cannot be initialized.
+    pub fn new() -> LnUrlHttpClientImpl {
+        let tokio_runtime = Runtime::new().expect("Failed to create Tokio runtime");
+        let tor_client = tokio_runtime.block_on(TorHttpClient::new()).unwrap();
+        LnUrlHttpClientImpl {
+            clearnet_client: reqwest::Client::new(),
+            tor_client,
+        }
+    }
```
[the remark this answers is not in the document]
Randy808: `address_filter` does not need to be marked as `mut` to contain a mutable reference.

See:
```
let mut a = A {x: 2};
// b is analogous to address_filter in the referenced code snippet (address_filter is of type '&mut ClientAddrConfigBuilder')
let b = &mut a;
b.x = 4;
println!("{}", b.x);
```

libs/gl-client/src/lnurl/pay/models.rs
Randy808: The goal here was not to make a full http client. There's no need for the use case of resolving lnurls. Maybe naming it `TorHttpClient` gave the wrong impression of the intended goal. 

All the examples of connecting a reqwest http client over Tor uses proxies. The arti crate [supports running a socks proxy](https://docs.rs/arti/latest/arti/#arti) but there were less examples for that so I opted to go for the higher-level arti-client crate which showed the client making [manual http requests in the docs](https://docs.rs/arti-client/0.10.0/arti_client/#example-making-connections-over-tor).

I'll defer the discussion to a later PR where I'll re-include the now-removed `tor_http_client`.

libs/gl-client/src/lnurl/tor_http_client.rs
Randy808: `address_filter` does not need to be marked as `mut` to contain a mutable reference.

See:
```
let mut a = A {x: 2};
// b is analogous to address_filter in the referenced code snippet (address_filter is of type '&mut ClientAddrConfigBuilder')
let b = &mut a;
b.x = 4;
println!("{}", b.x);
```

libs/gl-client/src/lnurl/tor_http_client.rs
Randy808: The goal here was not to make a full http client. There's no need for the use case of resolving lnurls. Maybe naming it `TorHttpClient` gave the wrong impression of the intended goal. 

All the examples of connecting a reqwest http client over Tor uses proxies. The arti crate [supports running a socks proxy](https://docs.rs/arti/latest/arti/#arti) but there were less examples for that so I opted to go for the higher-level arti-client crate which showed the client making [manual http requests in the docs](https://docs.rs/arti-client/0.10.0/arti_client/#example-making-connections-over-tor).

I'll defer the discussion to a later PR where I'll re-include the now-removed `tor_http_client`.

libs/gl-client/src/lnurl/mod.rs
cdecker: Discussed in-person: keeping it exploded for now, asking the user to ferry information back and forth, allowing to hook into the process makes sense. We can always build a `SimpleLnUrl` struct that wraps the node and the LNURL struct to simplify its use for users.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -1,25 +1,25 @@
-pub mod models;
-
+use super::models;
+use super::utils::parse_lnurl;
+use crate::lnurl::models::{
+    LnUrlHttpClient, MockLnUrlHttpClient, PayRequestCallbackResponse, PayRequestResponse,
+};
 use anyhow::{anyhow, ensure, Result};
-use bech32::FromBase32;
+use futures::future;
```
cdecker, line 7: This import looks unused? Well, it's `future::ready` is being used, but only in the `test` submodule as far as I can see.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -1,25 +1,25 @@
...
 use lightning_invoice::{Invoice, InvoiceDescription};
 use log::debug;
-use models::{
-    LnUrlHttpClient, MockLnUrlHttpClient, PayRequestCallbackResponse, PayRequestResponse,
-};
 use reqwest::Url;
 use sha256;
-use std::str::FromStr;
+use std::str::FromStr; // Import the future module
```
cdecker, line 12: The comment seems to have slipped between the `use` statements?

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -139,8 +128,18 @@ fn validate_invoice_from_callback_response(
...
+    use std::pin::Pin;
+
     use super::*;
 
+    fn convert_to_async_return_value<T: Send + 'static>(
+        value: T,
+    ) -> Pin<Box<dyn std::future::Future<Output = T> + Send>> {
+        let ready_future: Ready<_> = future::ready(value);
+        Pin::new(Box::new(ready_future)) as Pin<Box<dyn std::future::Future<Output = T> + Send>>
+    }
```
cdecker, line 141: Very clever, definitely something I need to remember :+1:

libs/gl-client/Cargo.toml
```diff
@@ -11,20 +11,26 @@ export = ["chacha20poly1305", "secp256k1"]
...
 log = "^0.4"
 pin-project = "*"
 prost = "0.11"
+reqwest = {version="0.11.18", features=["json"]}
 ring = "*"
 rustls-pemfile = "*"
+sha256 = "1.1.4"
 tokio = { version = "1", features = ["full"] }
 tonic = { version = "^0.8", features = ["tls", "transport"] }
+tor-rtcompat = {version="0.9.2"}
```
cdecker, line 29: Do we still use `tor-rtcompat` now that we removed `arti`?

cdecker (approved):
Great cleanup @Randy808 :+1:

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Removed

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Moved into the test mod

Randy808:
The build is failing after rebasing. I'm still looking into what it could be

libs/gl-client/Cargo.toml
```diff
@@ -11,20 +11,25 @@ export = ["chacha20poly1305", "secp256k1"]
...
+bech32 = "0.9.1"
 bytes = "1.2.1"
 hex = "*"
 http = "0.2"
 http-body = "^0.4"
+lightning-invoice = "0.23.0"
 log = "^0.4"
 pin-project = "*"
 prost = "0.11"
+reqwest = {version="0.11.18", features=["json"]}
```
cdecker, line 23: This is implicitly pulling in `openssl-sys` which depends on the openssl headers. We already use `rustls` through `tonic`, so we likely want to use the `rustls-tls-native-roots` feature.

cdecker (changes requested):
Adding `rustls-tls-native-roots` to the `reqwest` features frees us from the `openssl-dev` dependency

libs/gl-client/src/signer/mod.rs
```diff
@@ -679,7 +679,6 @@ mod tests {
                     signer_state: vec![],
                     requests: Vec::new(),
                 },
-                vec![],
```
Randy808, line 682: Removing this line since it's preventing the tests from building.

The method signature change was causing it to fail:
https://github.com/Blockstream/greenlight/pull/227/files#diff-8d5abc718d7a05d6432918246fcf5f9919681ca385b057f72a2117e9878e1932R284

cdecker (approved):

PR #230, opened 2023-08-07
Serialize auth blob

nepet:
Introduces a proto message to store the key and cert used for mTLS authentication, as well as a rune used for authorization in a common blob that can be used to create the tls identity from.
- [x] Add serialization
- [x] Add constructors for the node and the tlsConfig that use the auth blob
- [x] Pass through bindings
- [x] Tests
- [x] Add to `glcli` -> store and load it from disk.

This is based on #214 and is therefore set as draft. This will be finalized once #214 is merged.


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/230/commits/62eed13ee6da88b1ac66241a29622aa50ab71ed2

PR #232, opened 2023-08-15
Version 0.2

cdecker:
This is the preparatory branch for the next version of the client bindings. Adding it as a PR to make it more visible here.


nepet:
Rebased

nepet:
Rebased

cdecker:
I just pushed all crates and the python package, so we should soon be able to merge this, hopefully without breaking any users due to API changes.

nepet:
Rebased

cdecker:
Rebased on top of `main`, prepping all things for the switch to the
new API with the `AuthBlob`.

nepet:
Major re-org of the branch + renaming to `Credentials`.
Todo:
 - [ ]  Re-add documentation 
 - [ ] Re-add cli commands

libs/gl-client-py/src/credentials.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub fn new() -> Result<Self> {
+        Ok(Self {
+            typ: CredentialType::Nobody,
+            inner: credentials::Builder::nobody()?,
+            builder: credentials::Builder::new(),
+        })
+    }
+
+    #[staticmethod]
+    pub fn nobody() -> Result<Self> {
```
Randy808, line 36: Is this ever going to change from the implementation of 'new'? maybe we can just call new here so we only have to change the default params in one place

libs/gl-client/src/signer/mod.rs
```diff
@@ -247,19 +266,89 @@ impl Signer {
...
+        if let Some(device_id) = rune.get_id() {
+            checks.insert("".to_string(), format!("{}-{}", device_id, RUNE_VERSION));
+        }
+
+        // Check that the request points to `cln.Node`.
+        let mut parts = request.uri.split('/');
+        parts.next();
+        match parts.next() {
+            Some(service) => {
+                // Fixme: We skip this for now as we also have the
```
Randy808, line 305: What are we skipping and how would stricter validation look?

libs/gl-client/src/signer/mod.rs
```diff
@@ -713,14 +806,97 @@ impl Signer {
...
+    /// `create_rune("wjEjvKoFJToMLBv4QVbJpSbMoGFlnYVxs8yy40PIBgs9MC1nbDAmcHVia2V5PTAwMDAwMA", vec![vec!["method^list", "method^get"]])`
+    ///
+    pub fn create_rune(
+        &self,
+        rune: Option<&str>,
+        restrictions: Vec<Vec<&str>>,
+    ) -> Result<String, anyhow::Error> {
+        if let Some(rune) = rune {
+            // We got a rune, add restrictions to it!
+            let mut rune: Rune = Rune::from_base64(rune)?;
```
Randy808, line 862: It might be useful to do validity checks (like checking for pubkey field) here too since it doesn't look like [from_base64](https://github.com/nepet/futhark/blob/e659f4030a116867e1ecbb124342782b1f71cfbd/src/lib.rs#L707) checks for specific things. Maybe not a big deal but I'm thinking if there are more required restrictions it might be useful to consolidate the validity checks in a function that's called for both creating and modifying runes

libs/gl-client-py/src/credentials.rs
nepet: fc3e2f6a0524323edc7e4fdbef6e1fe70e63e6de

libs/gl-client/src/signer/mod.rs
nepet: Oh the comment seems to be off, actually we DO check the grpc service that was used to invoke the request.6ac1eb973ff4b60ce06436cb4fb2c171a9903bd2

libs/gl-client/src/signer/mod.rs
nepet: I initially implemented this check, but later removed it to avoid being patronizing towards developers who may want to create runes for their own use cases.

nepet:
A commit I added on top of this PR that changed the version numbers from `0.1.x` to `0.2.0` resulted in a CI failure with`cargo dist plan` so I removed it again.

While we could also add this later, it would still be nice to have a version update commit in here that we can point the backend to.

libs/gl-client-py/glclient/__init__.py
```diff
@@ -88,13 +89,13 @@ def export_node(self) -> schedpb.ExportNodeResponse:
         res = schedpb.ExportNodeResponse
         return res.FromString(bytes(self.inner.export_node()))
 
-    def node(self) -> "Node":
+    def node(self, creds: Credentials) -> "Node":
```
cdecker, line 95: We probably want to have `Scheduler` have a copy of the `Credentials`, since we will eventually enforce that only an authenticated client can schedule. That would mean that the `schedule()` call also needs the `Credentials`, and we could just inherit from `Scheduler` to `Node` when creating the `Node` here.

cdecker: Let's talk about this during our meeting.

nepet:
Added credentials and `upgradecreds` command to the command line tool.

cdecker (commented):
Excellent changes, I just bikeshedded the Scheduler and TlsConfig builder API a bit, but will merge this now, and we can iterate or leave things are as they are.

PR #234, opened 2023-08-17
Typehints for python grpc-methods

ErikDeSmedt:
I added typehints using the mypy-protobuf package.

This requries that `mypy-protobuf` python package is installed.

I created a `requirements-dev.txt`-file and added it to the docker-container. The bindings can now be regenerated as follows.

Do `make docker-shell`.

Inside the newly opened shell you can enter

```
make libs/gl-client-py/Makefile pygrpc
```

and

```
make libs/gl-testing/Makefile testgrpc
```


cdecker (approved):
We'll probably want to move the dependency list into `libs/gl-testing/pyproject.toml` or as `dev-dependencies` in `libs/gl-client-py/pyproject.toml`. Otherwise excellent change, thanks :+1:

PR #236, opened 2023-08-17
Removed spurious import

ErikDeSmedt:
I removed some unneeded imports which pulluted the logs with warnings

### DIFF
--- a/libs/gl-client/src/node/mod.rs
+++ b/libs/gl-client/src/node/mod.rs
@@ -1,5 +1,3 @@
-use std::time::Duration;
-
 use crate::pb::cln::node_client as cln_client;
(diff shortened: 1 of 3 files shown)


libs/gl-client/src/signer/.#mod.rs
```diff
@@ -0,0 +1 @@
+erik@erik-tp.561835:1692270444
```
cdecker, line 1: You probably don't want to commit this :-) But it's good to see a fellow emacs user :-)

cdecker: I think you might have forgotten to force-push, the PR still contains the file :-)

PR #237, opened 2023-08-17
signer: Add CLN v23.08 support

cdecker:
- [ ] Test that we handle 0-fee HTLCs correctly
 - [x] Test that we support 0-reserve + 0-fee channel type


cdecker:
Addressed the second open issue here: https://github.com/ElementsProject/lightning/pull/6586

PR #238, opened 2023-08-23
Install jq to fix make-clean

ErikDeSmedt:
If you run `make clean` inside a `make docker-shell` the operation fails. It appears `jq` is missing in the docker-image


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/238/commits/c4c3924db4cd9d67c9afb0039fefd8744afd065f

PR #245, opened 2023-08-29
Read from raw message

nepet:
We can leave out the bytes buffer in `process_request` altogether since we utilize the vls protocol message constructor `from_vec` that just takes the raw message.


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/245/commits/91487636e3104b98699a37b61cd19ee65cf215a6

PR #246, opened 2023-08-29
signer: Extract the bip32 key correctly

cdecker:
Turns out we did not parse the init message correctly, rather we used the trailer of the message which included the bip32 key, but also the fields after it.

@roeierez does this change impact any existing data or can this be switched in-place? If not we will add another method to retrieve the data formerly returned by `bip32_ext_key()` to maintain forward compatibility.

Reported-by: @roeierez


roeierez:
Thanks @cdecker 
Indeed we need the old format as we already have encrypted data (such as backup) that uses the previous key.

cdecker:
Thanks for the feedback, I will add a new method to retrieve the old bytestring. Any preference as what to call it? It will be in the API until anything that made use of the old behavior has been migrated. Is there a way for you to decrypt and re-encrypt the existing blobs to migrate away from the old behavior or are some of these on write-only media that can't be replaced with newer versions? In that case we'd keep it in there indefinitely until we're reasonably sure we'd not be breaking anything.

Apologies for the oversight, we were open-coding the parsing and made a wrong assumption (that the message would not be extended under our feet).

roeierez:
No worries, Christian, thanks for the support.
Currently only the sync.db that is stored in gl data store needs migration. My intention is to do that right on start of the sdk where we anyway check for new data. We will re-encrypt with the new key and that will be the only use of the old key.
We don't have a way to know that all nodes upgraded their backup but maybe after some time it would be reasonable to remove it considering the early stage we are now.

roeierez:
> Any preference as what to call it?

Should be something ugly the ring the bells. Not sure but maybe `legacy_bip32_ext_key` ? and mark as deprecated with some warning.

PR #249, opened 2023-08-31
signer: Implement the E2E verification

cdecker:
This is our initial implementation of the E2E verification.

# Overview

The E2E verification is used to ensure that any change the signer is being asked to sign off on originates from an authenticated client, i.e., anything that is initiated by an attacker or someone else who has access to the node host MUST NOT be signed off. This is done by the following chain:

 1. The client, using `gl-client` uses its local TLS identity to sign the grpc call payload
 2. The plugin extracts the signature and payload before decoding it (so we don't break signatures by having to re-encode) and stores them in a Context object.
 3. Upon receiving a signer request, the `gl-plugin` retrieves the current context and attaches it to the request
 4. Upon receiving an incoming request, the signer:
    1. Unpacks the context
    2. Verifies that the client TLS identity is indeed allowed to access the node (signer attestation sent along with the pubkey) for each of the payloads in the context
    3. Decodes the payloads into the corresponding grpc request messages
    4. Matches the request against the authenticated request messages, attempting to resolve which one caused the change
    5. If a resolution is found, process the signature request, if not, report a debugging warning, and reject the message
 5. The response is then sent the entire path back
 6. Eventually the grpc method call completes, at which point `gl-plugin` removes the corresponding payload from the context

This PR implements steps 4.IV and 4.V

# Reporting issues

The resolution of which grpc call caused the signer request is based on an allowlist implemented in `src/signer/resolve.rs`, and will reject any combination that wasn't explicitly allowed. As such it is bound to produce false negatives (rejections that should have been accepted and processed). Our goal is to incrementally refine the rules in `src/signer/resolve.rs` until we encounter no such false negatives.

For this purpose there is a debug information reporting facility built into the `gl-scheduler`, which allows us to collect failed requests and their context, and manually debug, check that it is indeed a false negative and adjust the resolution rules accordingly. The information shared in the report is already known to the server, but for simplicity we just forward it back, collated with the error message, allowing us to reconstruct the situation without having to combine multiple source of information.

 - [ ] Should we make that debug reporting optional?

In order to not disrupt the user-experience the rollout will happen in two steps:

 1. An initial permissive mode during which we collect debug information, add any missing resolution rules, and iterate on improving the signer
 2. An enforcement mode during which debug information is still being submitted, but any error is treated as fatal and will result in the message being dropped without processing any further.

Users can switch between enforcement and permissive mode using the `permissive` feature flag on the `gl-client` crate (default) or compiling with `--default-features=no` to enable enforcement. For `gl-client-py` use `maturin build --no-default-features` to enable enforcement.

When testing with `gl-testing` we will fail the `scheduler` fixture if any resolution failed, but tests run in permissive mode, since otherwise the way to see issue is by having the tests time out and then wading through hundreds of lines of backtraces.

# Status

We are actively testing these resolution rules and extending the server-side pipeline to facilitate fine-tuning. Feel free to try this out, however there may be some false negatives to deal with. Any feedback is of course very welcome :-)


libs/gl-client/src/signer/mod.rs
```diff
@@ -63,8 +68,14 @@ pub enum Error {
     #[error("scheduler returned faulty URI: {0}")]
     InvalidUri(#[from] InvalidUri),
 
-    #[error("signer refused a request: {0}")]
-    Signer(#[from] vls_protocol::Error),
+    #[error("resolver error: request {0:?}, context: {1:?}")]
```
nepet, line 71: I like this! I am far to sloppy, polluting everything with `anyhow!` errors. It is so much better to keep possible errors at one point in the file to find them!

libs/gl-testing/gltesting/fixtures.py
```diff
@@ -120,6 +121,15 @@ def mock_estimatesmartfee(r):
     del os.environ["GL_SCHEDULER_GRPC_URI"]
     s.stop()
 
+    # Check the reporting server to see if there were any failed
+    # request resolutions or policies. When testing we bail the test
+    # here.
+
+    if s.debugger.reports != []:
```
nepet, line 128: Calling a ValueError on teardown if something is unexpected is pure gold! Learned something new today!

libs/gl-testing/tests/test_node.py
```diff
@@ -205,7 +205,12 @@ def test_node_listpays_preimage(clients, node_factory, bitcoind):
         'preimage': preimage,
     })
 
+    from rich.rule import Rule
+    from rich.console import Console
+    console = Console()
+    console.rule("[bold red]<pay>")
```
nepet, line 211: I assume this is the sophisticated way of console.log debuging ^^.

Makefile
```diff
@@ -56,13 +56,12 @@ gen: ${GENALL}
 build-self: ensure-docker
 	(cd libs; cargo build --all)
 	(cd libs/gl-client-py && python3 -m maturin develop)
+	pip install -e libs/gl-testing
 
 check-self: ensure-docker
 	PYTHONPATH=/repo/libs/gl-testing \
 	pytest -vvv \
 	  /repo/libs/gl-testing \
```
nepet, line 64: Is there any specific reason why we shall remove the amount processes? Is this docker related?

Makefile
```diff
@@ -56,13 +56,12 @@ gen: ${GENALL}
 build-self: ensure-docker
 	(cd libs; cargo build --all)
 	(cd libs/gl-client-py && python3 -m maturin develop)
+	pip install -e libs/gl-testing
```
nepet, line 59: Super helpful! I always did this manually

libs/gl-client/src/signer/mod.rs
```diff
@@ -281,31 +289,25 @@ impl Signer {
...
-        use auth::Authorizer;
-        let auth = auth::GreenlightAuthorizer {};
-        let approvals = auth.authorize(ctxrequests).map_err(|e| Error::Auth(e))?;
-        debug!("Current approvals: {:?}", approvals);
-        Ok(approvals)
+        // Quick path out of here: we can't find a resolution for a
+        // request, then abort!
+        Resolver::try_resolve(msg, &reqs)?;
+
+        Ok(())
```
nepet, line 306: `Resolver::try_resolve` has the same return type as `authenticate_request` itself. We may obay returning an explicit `Ok(())` and instead return the result of `Resolve::try_resolve` directly.
```suggestion
        Resolver::try_resolve(msg, &reqs)?
```

nepet (commented):
I am really happy that this is moving forward and I like the newly introduced concepts (Resolver, Reporter, etc). My comments are for sure nitpicky!

libs/gl-client/src/signer/mod.rs
devrandom: ... without the `?` operator

libs/gl-client/src/signer/approver.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl<A: Approve> ReportingApprover<A> {
+    pub fn new(delegate: A) -> Self {
+        ReportingApprover { inner: delegate }
+    }
+}
+
+impl<A: Approve> Approve for ReportingApprover<A> {
+    fn approve_invoice(&self, inv: &lightning_signer::invoice::Invoice) -> bool {
+	log::warn!("unapproved invoice: {:?}", inv);
```
devrandom, line 22: nit: here and elsewhere - inconsistent tab vs spaces indentation

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+                (Message::SignMessage(m1), Request::SignMessage(m2)) => {
+                    m1.message.0.clone() == m2.message.as_bytes()
+                }
+                (Message::NewChannel(m1), Request::FundChannel(m2)) => {
+                    // Different node_id? Reject!
+                    m1.node_id.0 == m2.id.as_slice()
+                    // TODO: Add `close_to` to allowlist for the close
+                    // later on
+                }
+                (Message::NewChannel(m1), Request::GlFundChannel(m2)) => {
```
devrandom, line 76: may be good to also match `Message::ReadyChannel` so we can validate the amount and shutdown script

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+                }
+                (Message::SignInvoice(_l), Request::GlCreateInvoice(_r)) => true,
+                (Message::SignInvoice(_l), Request::Invoice(_r)) => {
+                    // TODO: This could be strengthened by parsing the
+                    // invoice from `l.u5bytes` and verify the
+                    // description, amount and (maybe) payment_hash
+                    true
+                }
+                (Message::PreapproveInvoice(l), Request::Pay(r)) => {
+                    l.invstring.0 == r.bolt11.as_bytes()
```
devrandom, line 88: I notice that the amount and description are "denormalized" in the request - i.e. they exist in both the bolt11 invoice and as top-level fields request.  is it assumed that a well behaved client would always keep these identical?

libs/gl-client/src/signer/resolve.rs
devrandom: also, when does the client generate a `GlFundChannel` and when a `FundChannel`?

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+            Message::Ping(_) => true,
+            Message::Pong(_) => true,
+            Message::SignChannelAnnouncement(_) => true,
+            Message::SignChannelUpdate(_) => true,
+            Message::SignNodeAnnouncement(_) => true,
+            // Duplicate verification with VLS, we defer to VLS
+            Message::GetChannelBasepoints(_) => true,
+            Message::ValidateCommitmentTx(_) => true,
+            Message::SignWithdrawal(_) => true,
+            Message::ReadyChannel(_) => true,
```
devrandom, line 33: see below about matching `ReadyChannel` against a request

devrandom (commented):
overall structure looks good

libs/gl-client/src/signer/mod.rs
nepet: Thanks @devrandom, changed the suggestion.

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+	    Message::SignAnyLocalHtlcTx(_) => true,
+	    Message::SignAnyRemoteHtlcToUs(_) => true,
+	    // Default to rejecting, punting the decision to the next
+	    // step.
+            _ => false,
+        };
+
+        // If we found a resolution, then there is no point in trying
+        // to match up further.
+        if accept {
```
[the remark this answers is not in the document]
cdecker: I'm afraid that won't work here: `reqctx` contains _all_ pending grpc commands, including read-only ones. So if we require this to be empty we may end up rejecting something that should be accepted.

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+            // Duplicate verification with VLS, we defer to VLS
+            Message::GetChannelBasepoints(_) => true,
+            Message::ValidateCommitmentTx(_) => true,
+            Message::SignWithdrawal(_) => true,
+            Message::ReadyChannel(_) => true,
+            Message::GetPerCommitmentPoint(_) => true,
+            Message::ValidateRevocation(_) => true,
+            Message::NewChannel(_) => true,
+	    Message::SignCommitmentTx(_) => true,
+	    Message::SignGossipMessage(_) => true,
```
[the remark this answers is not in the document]
cdecker: Correct, the resolution only looks for justification for user-triggered operations. Which excludes anything that can be triggered automatically (reconnect when the signer connects is triggered by `gl-plugin` for example). So this is _just_ the subset of operations that we may not require a user-action as justifciation.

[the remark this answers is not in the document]
cdecker: The VLS then does the actual bookkeeping, assuming the cause for the current request has been resolved, and deemed sufficient to be processed. This means we essentially prefilter whatever VLS gets to see, allowing us to keep VLS free from anything GL-specific, such as our authentication system, the method authorization, etc

libs/gl-client/src/signer/resolve.rs
```diff
@@ -0,0 +1,103 @@
...
+            );
+            return Ok(());
+        }
+
+        for cr in reqctx {
+            let accept = match (dbg!(req), dbg!(cr)) {
+                (Message::SignMessage(m1), Request::SignMessage(m2)) => {
+                    m1.message.0.clone() == m2.message.as_bytes()
+                }
+                (Message::NewChannel(m1), Request::FundChannel(m2)) => {
```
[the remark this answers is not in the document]
cdecker: You're right we probably aren't actually checking this combination because we allow all `NewChannel` call, so this is a bit superfluous. `NewChannel` also happens in case in the fundee, so I'm not sure we can guard it here.

libs/gl-client/src/signer/mod.rs
```diff
@@ -281,31 +289,25 @@ impl Signer {
...
-            .map(|r| decode_request(r))
-            .filter_map(|r| match r {
-                Ok(r) => Some(r),
-                Err(e) => {
-                    log::error!("Unable to decode request in context: {}", e);
-                    None
-                }
-            })
-            .collect::<Vec<model::Request>>();
+    fn authenticate_request(
```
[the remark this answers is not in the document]
cdecker: We may be pushing the context signature verification into it, but I'm still moving pieces around to see what works best.

libs/gl-testing/gltesting/fixtures.py
cdecker: Yeah, it's quite nice for invariant checking, and we do it a log in CLN itself. There is one caveat though: things like `flaky` will not rerun a test if it's a fixture that causes the error, rather than the test itself. So rerunning failures doesn't work for these post-mortem checks, and so we try to keep them to a minimum to avoid too much CI breakage.

Makefile
cdecker: Yeah, the scripts aren't quite there yet, but if we can make ours and our user's life easier we should do it.

Makefile
cdecker: I just don't like it too much on CI, because it actually slows CI down, and it has a tendency of messing up the stdout-capturing in pytest, causing a wall of text obscuring what we're actually looking at. CI is slow, but we don't really care too much, hence why running sequentially can be useful.

For my own testing I have a `pytest.ini` file in the repo root that has exactly the options I like when running tests locally.

libs/gl-client/src/signer/resolve.rs
cdecker: Good catch, I will add a match-arm for `ReadyChannel`.

The difference between `GlXyz` and `Xyz` is that one is the legacy `greenlight.proto` variant, and the non-prefixed one is the modern cln-grpc `node.proto` variant. We will be retiring the `greenlight.proto` version in v0.2 soon :tm:

PR #250, opened 2023-09-01
Added support for lightning addresses as per the lud16 spec.

Randy808:


libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -81,7 +131,7 @@ fn extract_description(lnurl_pay_request_response: &PayRequestResponse) -> Resul
         serde_json::from_str(&serialized_metadata.to_owned())
             .map_err(|e| anyhow!("Failed to deserialize metadata: {}", e))?;
 
-    for metadata in deserialized_metadata {
+    for metadata in deserialized_metadata.clone() {
```
cdecker, line 134: This seems superfluous given that we clone below anyway.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -123,6 +173,47 @@ fn ensure_amount_is_within_range(
...
+        r"^(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9][.]?$",
+    ) {
+        Ok(regex) => {
+            return regex.is_match(domain);
+        }
+        Err(_e) => {
+            return false;
+        }
+    }
+}
```
cdecker, line 215: I'm not sure we need an additional validation here: since we're using `reqwest` to parse the URL later, and it may still be happy with the `host` part of the URL, it should also be able to perform a DNS lookup for that host. Adding additional validation here seems more risky than warranted (reqwest will tell us if it was unable to resolve and contact a host anyway), and may introduce false negatives (UTF-8 domains are not accepted here, as are emojis, and I'm not quite convinced that IPv4 and IPv6 addresses would be accepted either). Please remove this, and the dependency on regex.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -123,6 +173,47 @@ fn ensure_amount_is_within_range(
     Ok(())
 }
 
+//LUD-16: Paying to static internet identifiers.
+pub fn parse_lightning_address(lightning_address: &str) -> Result<String> {
+    let lightning_address_components: Vec<&str> = lightning_address.split("@").collect();
```
cdecker, line 178: Add a check that this results in two parts. Abort if not.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -123,6 +173,47 @@ fn ensure_amount_is_within_range(
...
+
+    let username = lightning_address_components.get(0);
+
+    if username.is_none() {
+        return Err(anyhow!("Could not parse username in lightning address"));
+    }
```
cdecker, line 184: This should likely be `is_empty()` because `@mydomain.com` will produce a valid split, but the username is empty. Also without the 2 part check above, this will never get triggered:

```
mydomain.com -> ["mydomain.com"] -> valid username, but empty domain?
```

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -123,6 +173,47 @@ fn ensure_amount_is_within_range(
...
+
+    let path = username.unwrap();
+
+    let domain = lightning_address_components.get(1);
+
+    if domain.is_none() {
+        return Err(anyhow!("Could not parse domain in lightning address"));
+    }
```
cdecker, line 192: This is the error that'd be triggered by `mydomain.com` which is strange.

libs/gl-client/src/lnurl/pay/mod.rs
```diff
@@ -123,6 +173,47 @@ fn ensure_amount_is_within_range(
...
+
+    let domain = domain.unwrap();
```
cdecker, line 194: Patternmatching allows you not to use `unwrap()` in turn ensuring the check above and the unwrap don't get accidentally separated.

```
let domain = match domain {
  Some(d) => d.
  None => return Err(anyhow!("no domain"))
};
```

cdecker (changes requested):

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: This is intentional. UTF-8 characters in domains should *not* be accepted as there are ascii conversion standards for that. For example, the domain for '😂.com' would resolve to 'xn--g28h.com' which would pass validation here. Ip addresses are also not domains and aren't valid for lightning addresses so wouldn't be considered here either. 

That being said, I'll remove this function and leave validation/resolution for the reqwest library when actually making the request.

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Removed

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Removed

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Added the check

libs/gl-client/src/lnurl/pay/mod.rs
Randy808: Changed to use pattern matching

cdecker (approved):

PR #251, opened 2023-09-01
Updated the formatting of all rust files using 'cargo fmt'.

Randy808:
Formatting changes using 'cargo fmt'. I'll close the PR if you don't like the suggestions but it would be nice to standardize formatting.


cdecker:
I totally agree that we should get into the habit of formatting our code with `cargo fmt`, however it should likely be a gradual process, given that we have a couple of long-standing PRs open which would get broken by the reformatting. Shall we try to clean the PR queue, and then perform a lock-step formatting of the codebase? I have been trying to format files I touch in PRs in the hopes of slowly getting there, but it leaves some files lacking proper formatting for a long time.

Also, how do you guys feel about `pre-commit-hooks` that check for fomat lints, maybe run a couple of smoke-tests to ensure basic quality (and have CI then run the full suite)?

Ping @nepet @Randy808 @ErikDeSmedt :-)

nepet:
I definitely agree that we should use at least `cargo fmt`. 
I like the `pre-commit-hook` for formatting alone but would dislike to many things being performed on every commit that could also be done by CI. 

Also I tried to clean up `clippy` complaints and formatting issues numerous times in the past but I always failed because I tried to clean up too big chunks at once I guess. I find `clippy` pretty helpful to show unhandled parts (e.g. `write()`) and to get rid of  unnecessary allocations or clones. Currently it is just not that easy to use as it complains about many parts that are not related to the newly introduced changes. I would appreciate if we could somehow clean this up, step by step.

How about:
- Configure your IDEs to "format on safe" OR use `pre-commit-hook` for `cargo fmt`.
- Add `clippy` to the CI so that at least the errors are being addressed, we could suppress the warnings if this seems to be to much.

cdecker:
I absolutely agree with both comments:

 - Developer experience is important, and failing a CI job because of formatting is just plain bad, so ideally we'd add a process before checking in changes to fix up any malformed files
 - Slow git commits are terrible, which automatically means running tests at commit is not feasible, and autoformatting should only be done if the overall time is still sub 2s
 - Ideally format changes would only touch code that is being changed anyway, thus not detracting a reviewers attention during review.

I do believe that a personal formatting style is less of a problem, and as long as we have a common formatting, rather than the perfect formatting, means we are better off (see the `go fmt` discussion about it not being anybody's favorite format, but it is at least shared, and therefore recognizable by anybody).

I like @nepet's suggestion of using `pre-commit-hook` to apply `cargo fmt` to files that have been changed (it does so selectively, not on the entire codebase), as long as the pre-commit is fast enough (so let's not overload it too much). It is also best effort: a dev that hasn't installed the `pre-commit` script will simply not format their file, and the next dev that has it will rectify it (which goes slightly against the muddying the waters point above, but I think it's a good tradeoff between failing CI due to an unformatted change, and not doing any formatting).

Shall we give it a go? If yes, we need to:

 - [ ] Configure `pre-commit`
 - [ ] Write up the dev docs instructing potentialy developers to configure the hook
 - [ ] (optional) We could look for a Github action that autofixes and commits the fixes whenever someone opens a PR. This would not fail CI, rather fix, and formatting changes would be confined into a separate commit, rather than polluting the diff

cdecker:
Closing in favor of a gradual formatting in the hopes of not breaking PRs and feature branches :-)

PR #252, opened 2023-09-04
Add mypy-protobuf to the dev dependencies

nepet:
This PR adds the following two changes:
- Add `mypy-protobuf` to the `gl-client-py` dev dependencies so that we can compile the grpc files from a poetry shell.
- Improve the `Makefile` targets `clean-py` and `pygrpc` to allow them to be called from the top dir as well as from within `gl-client-py`


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/252/commits/6f15286f539b899555f58d5adca4407cef4af72e

PR #253, opened 2023-09-04
Runes follow-up

nepet:
Just some small additions. We remove the rune from the request signature as the rune is already bound to the public key by a restriction. There is no need to add additional bytes on every call.


nepet:
Closed this as I already included these changes in a `main-0.2` rebase. The two commits that should have been added here are a061b886270d642614d796ffbbc3ec31711f5299 and e340a7e02d249224d99fd44448336b143c741775

PR #256, opened 2023-09-05
Extend the Invoice request message of the plugin wrapper

### LINKED ISSUE #255: expiry dosn't affect the created invoice
When using the `invoice` gRPC method to create an invoice, passing `expiry` the returned invoice has the default expiry (1 week) ignoring the passed one.

nepet:
Due to greenlight.proto - which we need to retire at some point - we were missing some fields of the invoice rcp request message. This adds all fields that are set by cln_grpc and extends the conversion traits.

Resolves #255


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/256/commits/81b82a88d2e84742ae4aae73e6ea8fb47b2fd89c

PR #257, opened 2023-09-07
Add Rule sets for Rune creation

nepet:
The goal of this PR is to provide a system that makes it easy and secure to construct runes with certain restrictions. We provide rule sets for common restrictions and combinations of those.

Example from a user perspective:
**A user wants to create and hand out a rune to a bookkeeping application (we all have to do taxes eventually):**
The application should not be allowed to move funds, neither off-chain nor on-chain but shall be able to read information from the node. A coarse rule set that is easy to understand (or at least human readable) and fulfills the task would be a **readonly** set that can be used to construct a rune that restricts the owner to read methods.

Example from an application perspective (pairing-protocol)
**A Podcast application uses greenlight without a distinct singer. The application wants to be authorized for a direct debit of max 1000 sat per week**
During the pairing-process the application would request a rune that is restricted to pay max _amt_ sat per _delta_time_ via the rule **rate(amt, delta_time)** that the user can then confirm or reject.

Tasks:
- [x] Add rune factory
- [x] Add exemplary set of rules
  - [x] readonly
  - [x] pay
  - [x] disjunct set
- [x] Check runes on signer
- [ ] Bindings for the rules and the factory

Rules (incomplete, loose ideas):
- [ ] only off-chain
- [ ] only on-chain
- [ ] rate limiting (amt, time, amt/time)
- [ ] one-time-rune (n-times)
- [ ] time-restricted (until date)

We do not have to implement every rule from the beginning but I am starting with a hand full of rule sets and recipes to ensure that our approach is modular enough.

### DIFF
--- a/libs/gl-client-py/glclient/__init__.py
+++ b/libs/gl-client-py/glclient/__init__.py
@@ -180,6 +180,15 @@ class Node(object):
         return res.FromString(
             bytes(self.inner.call(uri, bytes(req)))
         )
+    
+    def list_peer_channels(self) -> clnpb.ListpeerchannelsResponse:
+        uri = "/cln.Node/ListPeerChannels"
+        req = clnpb.ListpeerchannelsRequest().SerializeToString()
+        res = clnpb.ListpeerchannelsResponse
+
+        return res.FromString(
+            bytes(self.inner.call(uri, bytes(req)))
+        )
 
     def list_closed_channels(self) -> clnpb.ListclosedchannelsResponse:
         uri = "/cln.Node/ListClosedChannels"
--- a/libs/gl-client/src/lib.rs
+++ b/libs/gl-client/src/lib.rs
@@ -69,3 +69,5 @@ pub use lightning_signer::lightning_invoice;
 
 pub(crate) const TCP_KEEPALIVE: Duration = Duration::from_secs(5);
 pub(crate) const TCP_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(90);
+
+pub mod runes;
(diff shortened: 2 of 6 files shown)


nepet:
Main tasks are done. 
We still need to define an initial signer context that we check against a rune
From CLN: ("method", "rate" (needs to persist), "time", "pnum", "pnameX", "parrN", "pubkey")

cdecker:
Rebased on top of  for merging.

cdecker:
Rebased on top of `main-0.2` for merging.

Randy808:
I like the idea but it's hard to see the changes introduced in this branch without a rebase

docs/src/reference/creds.md
```diff
@@ -0,0 +1,229 @@
...
+Greenlight utilizes various components, such as mTLS certificates and 
+runes, to provide secure access to a node. 
+Various pieces of data must be persisted and made available to the client 
+at runtime. Check out the [Security][security] page for more information.
+
+To simplify data management for users and developers, Greenlight uses 
+`Credentials` as a centralized location for the necessary information. 
+Credentials can be reconstructed in various ways and exported in 
+byte-encoded format for persistence.
+
```
ErikDeSmedt, line 12: Provide an introduction to the different type of credentials and when you want to use them.

Suggestion

```
The nobody-credential is not linked to a specific node. You can use this credential to `register` a new node or `recover` an existing one. We don't allow everyone to spin up an unlimited number of nodes. You can download your developer certificate on `https://greenlight.blockstream.com`.

A device credential can be used to connect and authenticate a device. You receive device-credentials when you `register` or `recover` a node. These credentials can be stored on disk or encoded as binary data.
```

Randy808 (dismissed):

cdecker:
I think this is good to go, but it needs fixing of the tests.
@nepet please fix them up and then we'll merge this asap, it's been pending for a while now.

Randy808 (approved):

PR #260, opened 2023-09-13
Implement JSON-rpc structs for LSPS2

ErikDeSmedt:
I have implemented the data-types that are required for LSPS2.

### DIFF
--- a/libs/gl-client/src/lsps/lsps0/common_schemas.rs
+++ b/libs/gl-client/src/lsps/lsps0/common_schemas.rs
@@ -144,6 +144,94 @@ impl<'de> Deserialize<'de> for MsatAmount {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct ShortChannelId {
+    scid: u64,
+}
+
+// constants for parsing of short_channel_id in bits
+const SCID_BLOCK_HEIGHT_BITSHIFT: u64 = 24 + 16;
+const SCID_TXID_BITSHIFT: u64 = 16;
+
+impl ShortChannelId {
+    pub fn new_from_u64(scid: u64) -> Self {
+        Self { scid }
+    }
+
+    // The scid or short channel id consits out of 8 bytes
+    //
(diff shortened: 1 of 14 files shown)


ErikDeSmedt:
When working further on this PR I've noticed some files became too bulky.
I am currently moving some code around which should make it easier to review

libs/gl-client/src/lsps/lsps0/common_schemas.rs
```diff
@@ -144,6 +144,94 @@ impl<'de> Deserialize<'de> for MsatAmount {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub struct ShortChannelId {
```
cdecker, line 148: We already have a pretty much identical type in cln-rpc: https://github.com/ElementsProject/lightning/blob/77e3c3a38e808986d4515d92302ed7666e8bc99a/cln-rpc/src/primitives.rs#L130-L183

Would it be possible not to duplicate that type? We also have an amount type, but I'm not sure it's stringified version matches what you need for the LSPS, so if that doesn't work out, that's ok. SCIDs on the other hand should be pretty much identical.

ErikDeSmedt: The class is equivalent and replacing it would be trivial. 
The gl-client is not depending on `cln-rpc` yet. 

If you confirm you want to include it as a dependency I'll make the change

cdecker: Yeah, you're right, adding a dependency on `cln-rpc` while clean, also means we pull in stuff like the UnixStream dependency which we definitely don't want in the client.

ErikDeSmedt: I've copied the ShortChannelId from cln-rpc.

If we ever split out the primitives we should be able to use them without creating a breaking change

cdecker (approved):

PR #268, opened 2023-09-20
signer: Reproduce the overflow issue

cdecker:
This branch uses a custom version of VLS (v0.10.0-rc1, with the to_pubkey policy commented out) so we can get to the actual issue without having to add the real secret seed into the test. It loads the inputs from the JSON file, uses the signer infrastructure to set up a similar environment, loads the state, and then feeds the failing request to the `Handler`.

It's rudimentary, but should allow us to test for regressions in VLS and gl-client much easier in future.

To run the test:

 - `cd libs/gl-client`
 - `cargo test`


cdecker:
Confirmed to be working by @ok300, thanks ^^

PR #270, opened 2023-09-20
Bugfix: json_rpc transport could ignore bytes

ErikDeSmedt:
I used the wrong method to write and read data from a cursors. In some cases bytes could be ignored which would result in bugs.

I've discovered this bug using clippy


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/270/commits/3205d0d5c69d7989d179e10e2c500935b21314cf

PR #277, opened 2023-09-26
js: Remove outdated bindings for the release

cdecker:
As discussed with @ErikDeSmedt, we don't want users to use the JS bindings for the time being. They are undocumented, and may not be fully functional at the moment. So remove them in order to preempt any misunderstanding.


ErikDeSmedt:
Ack

PR #278, opened 2023-09-26
Added sigint signal handler to glcli's hsmd command.

### LINKED ISSUE #205: glcli with stuck commands: ctrl+c does nothing
when pressing `ctrl+c` in a glicli command that is stuck, nothing happens. I need to use `kill [pid]` to actually `glcli`

Randy808:
Fixes #205

### DIFF
--- a/tools/glcli/glcli/cli.py
+++ b/tools/glcli/glcli/cli.py
@@ -16,6 +16,7 @@ import struct
 import sys
 import threading
 import time
+import signal
 
 
 logger = logging.getLogger("glcli")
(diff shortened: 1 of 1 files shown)


cdecker:
Good addition, but I wonder if it works. I thought that during FFI calls the signal handlers installed by the python runtime would not be called?

Randy808:
I think you might be thinking of the behavior of the replaced `run_in_foreground` call, which blocks the thread and doesn't return control. I changed it to `run_in_thread` to allow the current thread to wait for signals and close on sigint.

Instructions for testing:
-Comment out the scheduler's [`maybe_upgrade` call](https://github.com/Blockstream/greenlight/blob/03adfa85c3f2a2ae59681d312fdfc4c376639883/libs/gl-client/src/signer/mod.rs#L528) since it's not relevant to this PR and would causes my `glcli hsmd` to crash
-Rebuild gl-client-py `maturin build --strip --out ./dist`
-Reinstall gl-client-py whl `pip install ./dist/gl_client-0.1.9-cp37-abi3-manylinux_2_31_x86_64.whl --force-reinstall`
-`poetry build` the cli
-Reinstall `pip install dist/glcli-0.1.8-py3-none-any.whl --force-reinstall`
-Call `glcli hsmd`

cdecker:
That's what I get for reviewing half asleep. Good job :+1:

ACK https://github.com/Blockstream/greenlight/pull/278/commits/c10d8631eb5641b258150909ddf5863c1174eda7

PR #279, opened 2023-09-28
client: Retire redundant greenlight.proto methods

cdecker:
We are retiring the old, manually maintained, `greenlight.proto` `Node` service, with the exception of greenlight-specific methods, such as the HSM and the streaming methods. This should cut down on maintenance effort and be much clearer as to which method to use for what purpose. It's a bit of a breaking change, but any removed method already has their replacement ready in `node.proto`.

We switch to `node.proto` which is maintained alongside CLN itself, so it should be much more stable.


Randy808:
ACK 7ff1cea2a7ca8b96fb4297e9c8f72dde33de8b9f

cdecker:
Got the go-ahead from @roeierez, so we can merge these. Thanks for the confirmation @roeierez :+1:

PR #281, opened 2023-10-01
Signer-less Device Pairing

nepet:
This PR adds the "Pairing Process/Protocol" to the client stack. The Pairing-Process is a way to connect a signer-less client to a greenlight node. The pairing request has to be approved by a trusted device.

I'll add a more details about the process soon.


libs/gl-client/src/pairing/service.rs
Randy808, line 1: Is this an old iteration of attestation_device.rs?

libs/gl-client/src/pairing/attestation_device.rs
```diff
@@ -0,0 +1,265 @@
...
+use tonic::transport::Channel;
+
+type Result<T, E = super::Error> = core::result::Result<T, E>;
+
+pub struct Connected(PairingClient<Channel>);
+pub struct Unconnected();
+
+pub struct Client<T> {
+    inner: T,
+    tls: TlsConfig,
```
Randy808, line 30: Any strong objections to replacing `tls`, `rune`, and key with `creds`?

It would make the definition of the Client smaller and allows us to skip passing the node_id to `approve_pairing`

libs/gl-client/src/pairing/attestation_device.rs
```diff
@@ -0,0 +1,265 @@
...
+            .await?
+            .into_inner())
+    }
+
+    pub async fn approve_pairing(
+        &self,
+        session_id: &str,
+        node_id: &[u8],
+        device_name: &str,
+        restrs: &str,
```
Randy808, line 100: Can the restrictions argument here be dropped?

If we allow the attestation device to put whatever restrictions they want, the application requesting the pairing may too little access to fulfill their use-case. Can we make it like android permissions where the base permissions are all or nothing? A gl-client consumer can show the user the restrictions put on the app with the pairing request when they receive it in `get_pairing_data`. Then the pairing can be completed from the restrictions from the original pairing request on the server-side. 

Also, what do you think about letting the attestation device reject the pairing request? I'm thinking it might be helpful for the requesting application on the UX side if the user doesn't like the proposed restrictions

libs/gl-client/src/pairing/attestation_device.rs
```diff
@@ -0,0 +1,265 @@
...
+            })
+            .await?
+            .into_inner())
+    }
+
+    pub async fn approve_pairing(
+        &self,
+        session_id: &str,
+        node_id: &[u8],
+        device_name: &str,
```
Randy808, line 99: Can we make the `device_name` optional?

libs/gl-client/src/signer/mod.rs
```diff
@@ -761,16 +766,145 @@ impl Signer {
...
+
+    async fn process_pairing_approval(
+        &self,
+        req_id: u32,
+        req: ApprovePairingRequest,
+        stream: mpsc::Sender<SignerResponse>,
+    ) -> () {
+        let mut data = vec![];
+        data.put(req.session_id.as_bytes());
+        data.put_u64(req.timestamp);
```
Randy808, line 832: Should the timestamp be checked in this method to bound its staleness like what was mentioned here?:

https://github.com/Blockstream/greenlight/pull/281/files#diff-95a68fef05d112753f7fbc0859cb6df74b6df1b75994ceb8937170b228651c88R515

libs/gl-client/src/signer/mod.rs
```diff
@@ -761,16 +766,145 @@ impl Signer {
...
+        data.put(&req.node_id[..]);
```
Randy808, line 833: Could we pull the node_id from `self` here?

libs/gl-client/src/pairing/new_device.rs
```diff
@@ -0,0 +1,144 @@
...
+pub struct Client<T> {
+    inner: T,
+    uri: String,
+    tls: TlsConfig,
+}
+
+impl Client<Unconnected> {
+    pub fn new<T>(creds: T) -> Client<Unconnected>
+    where
+        T: TlsConfigProvider,
```
Randy808, line 25: Should we return an error if a `device_cert` is passed in?

libs/gl-client/.resources/proto/glclient/scheduler.proto
```diff
@@ -398,3 +425,134 @@ message SignerRejection {
...
+message ApprovePairingRequest {
+	string session_id = 1;
+
+	// The time that the old device approved the pairing request. This
+	// is used by the signer to restrict the duration an approval
+	// request is valid.
+	uint64 timestamp = 2;
+
+	// The id of the node to pair with.
+	bytes node_id = 3;
```
Randy808, line 518: Can we remove this since it looks like it's pulled from the cert on the server side?

libs/gl-client/src/signer/mod.rs
```diff
@@ -761,16 +766,145 @@ impl Signer {
...
+                debug!("Could not decode rune processing pairing approval {}", e);
+                return;
+            }
+        };
+
+        // Check that the rune matches
+        match self.verify_rune(PendingRequest {
+            request: vec![],
+            uri: "/cln.Node/ApprovePairing".to_string(),
+            signature: req.sig,
```
Randy808, line 857: Does the Context need to be updated to handle the signature?
https://github.com/Blockstream/greenlight/blob/e38a37613da7558c853f24be700c193f194a6bc9/libs/gl-client/src/runes.rs#L159

libs/gl-testing/gltesting/scheduler.py
```diff
@@ -413,11 +419,15 @@ async def PairDevice(self, req: schedpb.PairDeviceRequest):
             "restrs": req.restrs
         }
         self.sessions[req.session_id] = data
+
+        # Wait for the Approval of an old device.
+        await self.recv_once(self.recv_stream)
```
cdecker: nit: Not checking the type here is likely an error: you are assuming a canonical ordering on the events that come in through this stream, but the way it is constructed suggests that the intent is to add more use-cases to theis stream incrementally as we need them. Adding new messages could invalidate this assumption, and concurrent pairing requests will definitely cause desyncs. Guarding by looping until we get the correct message is the safer way of doing this, or fall back to classical call-response patterns where this association is not needed.

This is minor for now, but a potential source for headaches further down the road, and having an assert or something here, with a brief note about the assumptions could help a lot.

libs/gl-client/src/pairing/attestation_device.rs
```diff
@@ -148,4 +150,119 @@ impl Client<Connected> {
...
+
+            if pk == data.session_id
+                && Self::restriction_contains_pubkey_exactly_once(&data.restrs, &data.session_id)
+            {
+                Ok(())
+            } else {
+                Err(Error::VerifyPairingDataError(format!(
+                    "public key {} does not match pk {}",
+                    data.session_id, pk
+                )))
```
cdecker, line 176: This is a bit of circular message, can we come up with a more clear way of phrasing this? "The public key you asked us to sign does not match the session_id"?

libs/gl-client/src/pairing/attestation_device.rs
```diff
@@ -148,4 +150,119 @@ impl Client<Connected> {
...
+                if end_index < s.len() && s.chars().nth(end_index) == Some('|') {
+                    return false;
+                }
+
+                // Check if 'pubkey=<pubkey>' appears exactly once
+                s.matches(&search_field).count() == 1
+            }
+            None => false,
+        }
+    }
```
cdecker, line 208: Shouldn't we be checking for a standalone `pubkey=03....` restriction? That'd mean that there cannot be any qualifiers (alternatives) on that restriction, and any rune including this restriction has to adhere to that pubkey. Attempting to parse a logical expression, and testing it doesn't somehow negate the clause we're interested in is hard (NP-hard to be precise, even if parsing works perfectly).

libs/gl-client/src/pairing/service.rs
```diff
@@ -0,0 +1,345 @@
...
+                    .await?
+                    .into_inner())
+            } else {
+                Err(PairingError::ApproveParingError(
+                    "tls key missing".to_string(),
+                ))
+            }
+        } else {
+            Err(PairingError::ApproveParingError("rune missing".to_string()))
+        }
```
cdecker, line 275: Move this to the top:

```
if self.rune.is_none() {
    return Err(...)
}
// Continue with one fewer indendation.
```

The reason is that in long indented blocks are hard to see the matching indentations, and having to care about an early check at the end, when really it could just have been an early exit is just strange :-)

libs/gl-client/src/signer/mod.rs
```diff
@@ -761,16 +766,145 @@ impl Signer {
...
+            };
+        }
+    }
+
+    async fn process_pairing_approval(
+        &self,
+        req_id: u32,
+        req: ApprovePairingRequest,
+        stream: mpsc::Sender<SignerResponse>,
+    ) -> () {
```
cdecker, line 829: The empty return makes the following construction necessary:

```
debug!(...);
return;
```

Which could have been an inner function returning a real `Result<>`, thus only having to do the logging once, at a level decided once, and we could skip the `return`. Might be a bit more rustic.

cdecker (approved):
Quite an impressive PR! I went through and all I could find were small nits. There is one comment in there about the ordering assumptions on the streams, but that can be deferred until it becomes important, if we leave enough breadcrumbs to identify the issue once it arises, otherwise just naming suggestions, that you should feel free to ignore :-)

libs/gl-client/src/pairing/attestation_device.rs
nepet: The attestation side of the pairing process is stateless at the moment as in: We don't carry any application data form step to step but If it turns out that this would be easier for developers once people started to use it, I am absolutely happy to change this.

libs/gl-client/src/pairing/attestation_device.rs
nepet: To make the `device_name` optional we need to think about what would make for a good identifier such that we can revoke a single rune that we handed out to a device.

libs/gl-client/src/pairing/new_device.rs
```diff
@@ -95,13 +95,20 @@ impl Client<Connected> {
             // Step 1 of the Pairing Protocol: Request pairing at the Greenlight
             // Backend.
             let request = client.pair_device(PairDeviceRequest {
-                session_id,
+                session_id: session_id.clone(),
```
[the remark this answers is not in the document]
nepet: Unfortunately it seems that this is needed. We also need access to the session_id a few line below when constructing the "qr-data-string"

libs/gl-client/src/pairing/new_device.rs
nepet: Good question, what would you prefer?

libs/gl-client/src/signer/mod.rs
nepet: As the timestamp is part of the signature we also need to add it to the data in order to verify that the signature is correct.

cdecker:
ACK https://github.com/Blockstream/greenlight/pull/281/commits/f70230ffe096e62d8bc527a9297f9eca67b24208

PR #285, opened 2023-10-05
Update docs to reflect the new certificate flow

adi-shankara:
- remove all mentions of invite code
- introduce instructions to use the certificate under getting started
- minor fixes with syntax


docs/src/getting-started/certs.md
```diff
@@ -0,0 +1,58 @@
...
+
+Ideally the two files are then stored, securely, alongside the code,
+in encrypted form, or instrument your CI system to have access to them
+when building. Treat the private key with the same care you'd use for
+an API key, as they fulfill identical roles in this scenario.
+
+In order to tell the build to use the certificate you'll have
+to set two environment variables in the build environment. Please
+consult your build system and/or shell environment about how to ensure
+teh build sees the variables.
```
ErikDeSmedt, line 39: Typo, should be teh

docs/src/getting-started/register.md
```diff
@@ -148,7 +148,7 @@ to the [`Scheduler`][scheduler]:
 
 	// Passing in the signer is required because the client needs to prove
 	// ownership of the `node_id`
-	scheduler.register(&signer, Some(invite_code)).await.unwrap();
+	scheduler.register(&signer).await.unwrap();
```
ErikDeSmedt, line 151: This will not work. I haven't tested it yet but I think we'll need

````rust
scheduler.register(&signer, None).await.unwrap();
```

docs/mkdocs.yml
```diff
@@ -71,7 +72,6 @@ nav:
     - Overview: reference/index.md
     - Client Libraries: reference/client-libraries.md
     - Security: reference/security.md
-    - reference/partner-certs.md
```
cdecker, line 74: Don't rename the file, it'll leave any public links dangling. We could use something like `mkdocs-redirect` to put a redirection in place to make it work though

https://github.com/mkdocs/mkdocs-redirects

cdecker:
Very good, just two minor typos and a file being renamed that would leave incoming links dangling. Otherwise good to go :+1:

cdecker (changes requested):

cdecker:
I went ahead and set up the redirect. @adi-shankara once the typo and the missing Rust argument to `scheduler.register()` and we should be good to go.

cdecker (dismissed):

adi-shankara:
Typos are now fixed. Thanks for the reviews!

PR #289, opened 2023-10-09
Update to the latest `ring` version

cdecker:


cdecker:
Closing because MacOS builders seem not to like the new rustls version.

PR #300, opened 2023-10-22
plugin: Do not reconnect on each `pay`

cdecker:
Reconnecting can be racy, and may prolong the time until we have a
usable connection.

The problem is that reconnects force a reconnection on already
connected nodes too, which can then be racy with the `pay` itself
since `connected` will be `true` much earlier than the channel becomes
ready.

### DIFF
--- a/libs/gl-plugin/src/node/wrapper.rs
+++ b/libs/gl-plugin/src/node/wrapper.rs
@@ -298,7 +298,6 @@ impl Node for WrappedNodeServer {
     }
 
     async fn pay(&self, r: Request<pb::PayRequest>) -> Result<Response<pb::PayResponse>, Status> {
-        let _ = self.node_server.reconnect_peers().await;
         self.inner.pay(r).await
     }
 


libs/gl-plugin/src/node/wrapper.rs
```diff
@@ -293,7 +293,6 @@ impl Node for WrappedNodeServer {
     }
 
     async fn pay(&self, r: Request<pb::PayRequest>) -> Result<Response<pb::PayResponse>, Status> {
-        let _ = self.node_server.reconnect_peers().await;
```
roeierez, line 296: @cdecker  Will this still work if the user pays immediately after scheduling?

cdecker: It should: the signer connecting already triggers a reconnection, so this one is a bit redundant. And worse, if the signer connects, triggers a connection, then we follow with a `pay`, which reconnects, then we end up killing the first connection, and replace it with the second one, if the channel re-establish dance did not complete before `reconnect_peers()` was called. In the logs this can be seen by repeated `WIRE_INIT` coming in at startup followed by a `pay`.

roeierez: Sounds like it is an improvements. I just wonder if we still have a race where we pay immediately after schedule and the channel is not active yet (even though the connect was triggered).
The signer starts asynchronously and paying immediately might hit the pay method before the channels is active.
Perhaps a good way to make the pay solid here is to replace the reconnect_peers by waiting for at least one channel to be active (assuming there is at least one persistent peer)?
Just adding this as something to think of as we did that in breez mobile and it served this purpose very well.

cdecker: That is correct, we are indeed hitting the situation where we did reconnect, but the channel re-establishment is still ongoing when the first HTLC hits. This causes `lightningd` to report `first peer not ready`, because the re-establish flow did not complete. I am looking for a reliable signal that indicates the re-establish flow is finished and the HTLC can go through.

FWIW that is the reason why payments to WOS on https://pulse.btxtb.de/d/d60d5e32-ef33-4b95-ab43-4137e3b64209/pulse reliably fail, as it is the first of the three test payments, and therefore the one that runs into this issue.

PR #305, opened 2023-10-26
Identity from path

ErikDeSmedt:
Implements `TlsConfig::identity_from_path`.

This is a utility method that reads an identity (cert + key) from a path.

The python implementation might be a bit controversial. I have removed the `id` which is a breaking change.

The id is the combination of `cert` + `key` and is only used in the python version. When using `identity_from_path` there are no python bindings available to request the `cert` + `key` and construct the `id` gain.

The alternative is to read the `client.crt` and `client-key.pem` file in python and use the standard `tls.identity`-call. However, I prefer not to duplicate the code.


nepet (approved):
ACK 709a76413e409b7a4a14b78777f6df398b695400

Was too fast! Seems that gl-testing relies on the `id` field!

libs/gl-client-py/glclient/tls.py
```diff
@@ -11,7 +11,6 @@ def __init__(self):
         # through the rust library (streaming calls)
         self.inner = native.TlsConfig()
         self.ca: Optional[bytes] = None
-        self.id: Tuple[Optional[bytes], Optional[bytes]] = (None, None)
```
nepet, line 14: This id field is accessed by some tests from `gltesting/tests/test_gl_node.py` They are necessary there in order to set up a custom gprc connection. I thing the custom grpc connection is not needed anymore and actually not the right way to access greenlight.
I can add a fix on top of your commits if you like, as I wrote the tests that need the id field in the first place.

ErikDeSmedt: That would be great

nepet (approved):
Added the test fixes on top and rebased on main

PR #312, opened 2023-11-02
signer: Extract invoices from `sendpay` command and add them to the known invoices

cdecker:
This is necessary for `sendpay` to work correctly, i.e., have the signer accept a change in balance that results from a `sendpay`, and not complain about the unexplained balance change.

Notice that this adds invoices with whicht the `sendpay` is annotated with, but does not yet register a spontaneous payment if no invoice is attached.

Reported-by: @roeierez

### DIFF
--- a/libs/gl-client/src/signer/mod.rs
+++ b/libs/gl-client/src/signer/mod.rs
@@ -398,7 +398,7 @@ impl Signer {
 
         use auth::Authorizer;
         let auth = auth::GreenlightAuthorizer {};
-        let approvals = auth.authorize(ctxrequests).map_err(|e| Error::Auth(e))?;
+        let approvals = auth.authorize(&ctxrequests).map_err(|e| Error::Auth(e))?;
         debug!("Current approvals: {:?}", approvals);
 
         let approver = Arc::new(MemoApprover::new(approver::ReportingApprover::new(
--- a/libs/gl-client/src/signer/mod.rs
+++ b/libs/gl-client/src/signer/mod.rs
@@ -410,8 +410,12 @@ impl Signer {
         approver.approve(approvals);
         let root_handler = self.handler_with_approver(approver)?;
 
-        // Match over root and client handler.
+        log::trace!("Updating state from context");
+        update_state_from_context(&ctxrequests, &root_handler)
+            .expect("Updating state from context requests");
+        log::trace!("State updated");
 
+        // Match over root and client handler.
         let response = match req.context {
             Some(HsmRequestContext { dbid: 0, .. }) | None => {
                 // This is the main daemon talking to us.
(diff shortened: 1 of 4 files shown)


libs/gl-client/src/signer/mod.rs
```diff
@@ -681,6 +685,46 @@ impl Signer {
...
 
+/// Look through the context requests and update the state
+/// accordingly. This is useful to modify allowlists and invoice lists
+/// extracted from the authenticated requests.
+fn update_state_from_context(
+    requests: &Vec<model::Request>,
+    handler: &handler::RootHandler,
+) -> Result<(), Error> {
+    log::debug!("Updating state from {} context request", requests.len());
+    let node = dbg!(handler.node());
```
nepet, line 696:
```suggestion
    let node = handler.node();
```

libs/gl-testing/Dockerfile
```diff
@@ -109,12 +106,17 @@ RUN apt-get update && apt-get install -qqy \
...
     socat \
     unzip \
     wget \
     jq \
     && rm -rf /var/lib/apt/lists/*
 
+RUN groupadd -g $GID -o $DOCKER_USER &&\
+    useradd -m -u $UID -g $GID -G sudo -o -s /bin/bash $DOCKER_USER && \
+    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
```
nepet, line 119: As commands in the docker container did not fail to me before, and because I can not find make recipes or commands that use sudo inside of the docker container, I am just wondering why we need to add the user to the sudoers here?

libs/gl-testing/tests/test_node.py
```diff
@@ -425,3 +425,67 @@ def test_vls_crash_repro(
...
+
+    res = clnpb.SendpayResponse.FromString(
+        bytes(gl1.inner.call("/cln.Node/SendPay", req.SerializeToString()))
+    )
+
+    req = clnpb.WaitsendpayRequest(payment_hash=payment_hash)
+    res = clnpb.WaitsendpayResponse.FromString(
+        bytes(gl1.inner.call("/cln.Node/WaitSendPay", req.SerializeToString()))
+    )
+    print(res)
```
nepet, line 491: Maybe we want to assert a response here?

nepet (commented):
I really like the request/context flow here!

libs/gl-testing/Dockerfile
cdecker: I often end up wanting some more tools, for example `lnav` is very useful to drill down into the logs. Without `sudo` we can't install additional packages into the docker image, as frowned-upon as that may be, I really like having the ability to pull in more tools in the `docker-shell`.

libs/gl-testing/tests/test_node.py
cdecker: We'd be raising an issue anyway if this is not a valid response, so just calling is sufficient already.

PR #314, opened 2023-11-02
Add rustfmt warning to CI

nepet:
This adds a job to the CI that checks every rust file that is changed by the PR for formatting. It will however not fail the CI (`continue-on-error: true`) and still result in a green check mark.

We could add `check-rustfmt` as a prerequisite to `check-self` if we want to run it locally and fail the CI if the formatter complains on changed files

Another option would be to use something like https://github.com/marketplace/actions/continue-on-error-comment to write a comment into the PR about formatting issues. That would not be as aggressive as to fail the CI or to enforce local checks.

Tell me what you think about this!

### DIFF
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -1,0 +1,28 @@
+name: Check Formatting
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - opened
+  workflow_dispatch:
+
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/314/commits/5c556bd6bb591ab9579f38847c8aba2e246f1fff

PR #333, opened 2023-11-27
rs: allow `FetchInvoice` requests to the signer

hydra-yse:
When calling the `FetchInvoice` command, incoming requests passing through the signer currently seem to fail with the error `Unknown URI {}, can't decode payload`, due to not being exposed in the signer. This small PR aims to fix this


cdecker:
Thanks @hydra-yse, while we do complain loudly about being unable to decode the message, we should not be aborting the resolution due to it. This is exactly to allow us to backfill while collecting the errors, without breaking existing signers.

Thanks for adding these details.

PR #338, opened 2023-12-08
testing: Update pyln-testing dependency to match node version

cdecker:
v23.11 was released and the `--developer` flag no longer exists. We also now have `clnrest.py` to contend with.


cdecker:
Turns out that I had the change backwards, we need to pin to 23.08 until pyln-testing learns to identify the client version.

PR #344, opened 2023-12-16
Signer loop: retry upgrade on connection error

ok300:
If the signer loop (`run_forever`) is called when there is no network available, the loop exits because `scheduler.maybe_upgrade` fails.

We've tried handling this outside the signer loop, by re-starting the loop when connection-related errors are thrown (https://github.com/breez/breez-sdk/pull/691). But it's rather unwieldy and verbose.

Handling it inside the signer loop seems like a better choice, which is the focus of this PR.


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/344/commits/513cfc03c490f4ce859b6019e91bba5f8dff21c2

Issue #276, opened 2023-09-26
signer: Improve the E2E verification


cdecker:
There are a number of places where we can strengthen the existing E2E verification, or fine-tune to allow a more complete representation of the RPC semantics, e.g., by adding additional call parameters to be extracted and reflected in approvals.

 - [ ] @devrandom points out that we can validate the amount and shutdown script for `ReadyChannel` (https://github.com/Blockstream/greenlight/pull/249#discussion_r1314229879)
 - [ ] We need to add the `maxfeepercent` to the `pay` method to adjust the fees that we are willing to pay.
 - [ ] @devrandom points out that there may be overpayments for invoices, so we should extract the actual amount from the `preapproveinvoice` message (https://github.com/Blockstream/greenlight/pull/249#discussion_r1314230493)

PR #346, opened 2024-01-03
Upgrade VLS to v0.11.0rc1

cdecker:
This is in preparation for the CLN v23.11 deployment, and addition of
splice support. Splicing operations may not yet be exposed via grpc,
but we need the signer to support them first anyway.

One question for @devrandom is whether the newly introduced `SignerId`
is safe to set to a constant in our use-case, since signers are
intended to be clones of each other.

### DIFF
--- a/libs/gl-client/src/signer/resolve.rs
+++ b/libs/gl-client/src/signer/resolve.rs
@@ -32,7 +32,7 @@ impl Resolver {
             Message::GetChannelBasepoints(_) => true,
             Message::ValidateCommitmentTx(_) => true,
             Message::SignWithdrawal(_) => true,
-            Message::ReadyChannel(_) => true,
+            Message::SetupChannel(_) => true,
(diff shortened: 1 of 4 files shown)


libs/gl-client/src/signer/resolve.rs
```diff
@@ -32,7 +32,7 @@ impl Resolver {
             Message::GetChannelBasepoints(_) => true,
             Message::ValidateCommitmentTx(_) => true,
             Message::SignWithdrawal(_) => true,
-            Message::ReadyChannel(_) => true,
+            Message::SetupChannel(_) => true,
```
Randy808, line 35: We probably want to update issue #276 which refers to the outdated `ReadyChannel` when this merges

cdecker: Excellent catch :+1:

PR #350, opened 2024-01-26
signer: Add preapproveinvoice for client-driven payments

cdecker:
Manually sending `preapproveinvoice` was failing to adjust
the context on the signer, which then declined any subsequent HTLC
matching the invoice.

Reported-By: Domenico Gabriele


Randy808 (approved):

nepet (approved):
Ack 270e1ee812ecff757170a27d13b9ad98382069eb

Issue #335, opened 2023-11-29
Multiple versions of `bitcoin` dependency


ok300:
There are 3 versions of the `bitcoin` crate being used:

https://github.com/Blockstream/greenlight/blob/0e0ec904b6b84bbf89e3e59a2cd57f86538c71f4/Cargo.lock#L264-L265

https://github.com/Blockstream/greenlight/blob/0e0ec904b6b84bbf89e3e59a2cd57f86538c71f4/Cargo.lock#L277-L278

https://github.com/Blockstream/greenlight/blob/0e0ec904b6b84bbf89e3e59a2cd57f86538c71f4/Cargo.lock#L291-L292

Consolidating on one would remove bloat in the dependency tree, plus reduce binary size, which is a big plus for mobile clients.

PR #352, opened 2024-02-07
Remove duplicate bitcoin dependency

ok300:
This PR replaces the explicit `bitcoin` dependency with the `lightning_signer::bitcoin` re-exported one, for two reasons:

First, the explicit one it's not used. The re-exported `lightning_signer::bitcoin` is already used as such, for example

https://github.com/Blockstream/greenlight/blob/62ade9c53b68b8f7625725033121babb51720dbb/libs/gl-client/src/export.rs#L5-L8

Second, the explicit dependency brings with it a separate version into the dependency tree. Removing it and instead re-using the exported one consolidates the dependency tree and helps remove duplicates.

For example, before this PR, GL had 3 separate `bitcoin` versions in the dependency tree.

```bash
cargo tree --duplicate | grep bitcoin

...
bitcoin v0.29.2 (*)
bitcoin v0.30.2 (*)
bitcoin v0.31.0 (*)
```

With this PR, there are only 2 left:

```bash
cargo tree --duplicate | grep bitcoin

...
bitcoin v0.29.2 (*)
bitcoin v0.30.2 (*)
```

Related to https://github.com/Blockstream/greenlight/issues/335

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,12 +249,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
-name = "bech32"
-version = "0.10.0-beta"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ok300:
This also consolidates a few other duplicate transitive dependencies, for example `bech32`, `bitcoin_hashes` and `secp256k1` (see `Cargo.lock` PR changes).

PR #368, opened 2024-02-28
Expose `SignerHandle` in gl-client-py

ErikDeSmedt:
The `SignerHandle` isn't exposed in the native-bindings. However, the `glclient.pyi` file explicitly mentions the handle should be their.


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/368/commits/2851f26b217ac017d3872eb68d0af1bd97a8d010

PR #369, opened 2024-02-28
Improve build process of grpc bindings

ErikDeSmedt:
We are generating/compiling python bindings from `.proto`-files.
    
    Before this commit the generated protobuf-files did not understand they
    were a part of the `glclient`-module.
    
    This resulted in broken imports which we fixed using `sed` in our
    Makefile.
    
    This commit improves the configuration and voids the need for hacks in
    our `Makefile`.
    
    For more info: See https://github.com/protocolbuffers/protobuf/issues/4176


cdecker:
Following up on our discussion re Google path semantics, I took a look at the generated `_grpc.py` files, and was surprised the URI was unchanged, instead its just the prefix for the serialization and deserialization was changed. This makes my concerns from yesterday invalid ^^


Today I learned :+1:

cdecker:
Rebased to remove that merge commit, merging asap

PR #372, opened 2024-02-29
signer: Upgrade to VLS v0.11.0-rc2

### LINKED ISSUE #348: Withdraw results in `non-mandatory-script-verify-flag (Witness program hash mismatch). Unsent tx discarded`
A user is reporting several failed attempts to withdraw funds after a force close. The failure always results in the error:

`
Error calling method Withdraw: RpcError{ code: Some(-1), message: \"Error broadcasting transaction: error-code: -26\\\\nerror message: \\\\nnon-mandatory-script-verify-flag (Witness program hash mismatch). Unsent tx discarded ...
`

after which the utxo's are marked reserved for a day or so. Once the utxo's are available again, the user repeats resulting in the same error. Node id for reference is: 02810d14934b1b41b52af4b6c118878a00e857cc4922442ea59e18b31906b09c67

A similar issue is also raised in https://github.com/Blockstream/green_android/issues/185 relating to the same circumstances and error.

I have requested the user provides a trace log the next time they attempt a withdraw and will attach them to this issue.

cdecker:
Fixes #348


devrandom:
ACK from VLS point of view

PR #376, opened 2024-03-03
Removed network from Node implementation.

Randy808:
The passing around of the network argument can cause confusion amongst new developers since a node is registered with a single network and can't be dynamically changed post-registration. This draft PR proposes the removal of the network argument to better reflect the fact that a node's network is static after it has been registered.


ErikDeSmedt:
I like the idea of changing the API so users cannot shoot themselves in the foot.

Is it possible to register two nodes with the same `node_id` on a different network?
Are there any old nodes like this?

Once the python API is finished I'd be happy to approve this PR

Randy808:
Currently it's not possible to register 2 nodes with the same node_id on different networks and there are no greenlight nodes in this situation

Randy808:
Just updated Python interface

cdecker:
Hm, I thought I had replied to this one, but it seems to have been lost.

Yes, we do not allow registering the same node_id multiple times, even if they were on separate networks. This reflects the common best practice to not reuse secret keys in multiple applications, that while secure individually, may leak information if used in combination.

(also, it's a unique key in the database, simplifying queries :-p)

ErikDeSmedt (approved):

PR #383, opened 2024-03-07
Replace TlsConfig

nepet:
This PR replaces `TlsConfig` in the signer and the scheduler completely. Also removes the exaggerated credentials builder in favor of a flattened api.

@cdecker: Please tell me what you think about 744e282 - *scheduler: Refactor scheduler to use typestate*. This commit provides a more idiomatic (rust) way to handle the scheduler and the credentials. If we don't like it we can just strip it off before we merge, the commit below is in a working state as well.

After we decided if we want to go with this change I need to adapt the backend tests (they use glclient) before we can merge.


libs/gl-client-py/src/credentials.rs
```diff
@@ -1,90 +1,145 @@
...
+
+    pub fn is_device(&self) -> Result<()> {
+        if let Self::Device(_) = self {
+            Ok(())
+        } else {
+            Err(credentials::Error::IsIdentityError(
+                "credentials are not of type device".to_string(),
+            ))?
+        }
+    }
```
cdecker, line 44: The `is_` prefix suggested to me that this was a `bool`-return, when really it is supposed to be used in combination with `?` to provide an abort criterium. We probably want to rename this to `ensure_device()` as it does just check.

cdecker (dismissed):
Very nice, just a minor naming bikeshedding comment.

It's too bad that adding the typestate pattern in rust requires unifying shims again on the python side. Once we switch to uniffi, we might have to backport some of those shims into gl-client itself (and behind a feature-flag so only bindings get that code) so we can share it among all bindings.

libs/gl-client-py/src/credentials.rs
nepet: changed `is_` to `ensure_`

cdecker:
I like the idea of rewriting it in a language that is harder to drift apart, and it allows us to dogfood the rust bindings rather than the less strict python bindings 👍

nepet:
@Randy808 Seems that we have some conflicting commits. I removed the `Result<>` from the `TlsConfig` as it was not used before. This way I can create defaults that get checked during connect. I like the idea of a validation during construction as well. What do you think, should we bubble up the result to the surface layer during construction of the Credentials (unfortunately, this way we won't be able to create defaults anymore) or should we panic on the creation of `TlsConfig` or creation of the `Credentials` which I guess is okay as we only run on client side but maybe developers would dislike it? Which way would you prefer?

libs/gl-client-py/src/scheduler.rs
```diff
@@ -1,144 +1,264 @@
...
+        &self,
+        signer: &gl_client::signer::Signer,
+    ) -> Result<pb::scheduler::RecoveryResponse> {
+        match self {
+            UnifiedScheduler::Unauthenticated(u) => u.recover(&signer).await,
+            UnifiedScheduler::Authenticated(a) => a.recover(&signer).await,
+        }
+    }
+
+    async fn authenticate(&self, creds: R) -> Result<Self> {
```
Randy808, line 57: Could we constrain this to return an `Ok` when unauthenticated and an `Err` otherwise? I feel like preventing `Authenticated -> Authenticated` would help discourage instances of UnifiedSchedulers from being used as re-usable builders. Would it also be possible to pass `self` instead of `&self` to help enforce this?

Randy808:
@nepet In the TlsConfig's [`new`](https://github.com/Blockstream/greenlight/blob/main/libs/gl-client/src/tls.rs#L53) you can change the match to return an `Option` instead of returning an error on failure. That way we can keep changes to a minimum since I also like the defaults.

It would also stay consistent with the pattern followed for [`load_file_or_default`](https://github.com/Blockstream/greenlight/blob/92cd8aa72b7df028cf0db0dd7f20fcf2eb4c308e/libs/gl-client/src/tls.rs#L41) and what I did for [`identity`](https://github.com/Blockstream/greenlight/blob/cf1ca284b82908243db01eaa3e4c8c3df345a91c/libs/gl-client/src/tls.rs#L86) (since I didn't want to panic there for the same reasons)

libs/gl-client-py/src/scheduler.rs
nepet: It would certainly be possible to consume the scheduler here but unfortunately we can not consume the scheduler in a pyo3 class, so this wouldn't help with the public api. Let me just add an `Err` if the scheduler is already authenticated. I guess we should do the same in the rust library then?

libs/gl-client-py/src/credentials.rs
```diff
@@ -1,147 +1,145 @@
...
-        NobodyBuilder::new()
+    pub fn from_bytes(data: &[u8]) -> Self {
+        let inner = UnifiedCredentials::Device(gl_client::credentials::Device::from_bytes(data));
+        Self { inner }
     }
 
-    pub fn tls_config(&self) -> Result<TlsConfig> {
-        let inner = self.inner.tls_config()?;
-        Ok(TlsConfig { inner })
+    pub fn upgrade(&self, scheduler: &Scheduler, signer: &Signer) -> Result<Credentials> {
```
ErikDeSmedt, line 107: I'm a bit confused here. 

What is the process of upgrading `credentials` doing?
Why would I want to `upgrade` my credentials?

ErikDeSmedt (commented):
I like this set of changes

libs/gl-client-py/src/credentials.rs
nepet: This method exists to give us some compatibility on future changes. If the binary format of the credentials change, or if we add new fields, we have a function that can take care of reflecting these changes. Right now what it does is to call `recover` on the scheduler service to get a full set of credentials containing a rune if we don't have a rune yet.

Randy808 (approved):

PR #384, opened 2024-03-07
Improve poetry file

ErikDeSmedt:
A general clean-up of the `pyproject.toml`-files.
It is now possible to install all dependencies using `poetry install`.

Before you had to do
```
cd docs; poetry install
cd libs/gl-client-py; poetry install
cd libs/gl-testing; poetry install
```

Another advantage is that it now properly resolves dependencies and
conflicts. In the old approach conflicts between two different
`pyproject.toml`-files wouldn't be detected.

Users who want to contribute to the documentation have to use

```
poetry install --with docs
```


libs/gl-testing/pyproject.toml
```diff
@@ -30,7 +30,17 @@ anyio = "^3.7.1"
 purerpc = { version = "^0.8.0", extras = ["grpc"] }
 pytest-sugar = "^0.9.7"
 
-[tool.poetry.dev-dependencies]
```
ErikDeSmedt, line 33: This name has been deprecated

nepet:
In case anybody is getting an error similar to `'name'` when running `poetry shell` -> updating poetry to 1.8.2 helped: `poetry self update`

nepet (approved):
ack bddb115b8b4df1dbe4ddd45af3b7910c07212bf8

cdecker:
Good proposal, we should definitely have a top-level `pyproject.toml` file, that then pulls in the direct dependencies only, and the sub-projects that the main project requires. I don't like pulling in the docs dependencies into the main dependencies too much.

We could instead of moving the doc-dependencies into the main `pyproject.toml` file, rather add a dependency on `docs (and rename it to something that doesn't clash with PyPI such as `gl-internal-docs`) and then that one should have all its dependencies.

ErikDeSmedt:
> I don't like pulling in the docs dependencies into the main dependencies too much.

Initially, I followed your sentiment but I realized the maintenance burden is just to high.

I see two reasons why you would want to create a separate `pyproject.toml`-file
1. You want to do `cd docs; poetry install` and have a working environment
2. You want to package the project

We want to package and publish `gl-client` and `gl-testing` thus we need a `pyproject.toml`-file.
We don't want to package the docs so we don't need a `pyproject.toml`-file.

## How it increases maintenance burden?

In `poetry` dev-dependencies are local to a single project and poetry never installs dev-dependencies from a dependency.

If we want to add a type-stub as a dev-dependency to `gl-client-py` we would have to add it to

- `./libs/gl-client-py/pyproject.toml`
  Because `cd ./libs/gl-client-py; poetry install` should install the dependency
- `./pyproject.toml`
  Because `poetry install` should install the dependency
- `./docs/pyproject.toml`
  Because `cd docs; poetry install` should install the dependency. We should build the [reference](https://blockstream.github.io/greenlight/py/glclient.html) using typestubs

I don't want to create a `pyproject.toml`-file that we'll never use and that get's outdated easily.

## Groups are recommended

The first paragraph in the [docs](https://python-poetry.org/docs/managing-dependencies/) about dependency management is 

> Poetry provides a way to organize your dependencies by groups. For instance, you might have dependencies that are only needed to test your project or to build the documentation.

cdecker:
Excellent rationale, thank you. I think I had not fully understood the intent and now that I do I fully agree 👍

PR #385, opened 2024-03-07
Drop support for npn_protocols

ErikDeSmedt:
`set_npn_protocols` has been removed in python 3.11. 

We are already using `set_alpn_protocols` so I think we can drop it safely.

This should help us to run `gl-testing` in standalone mode

### DIFF
--- a/libs/gl-testing/gltesting/identity.py
+++ b/libs/gl-testing/gltesting/identity.py
@@ -59,10 +59,6 @@ class Identity(object):
         s.load_cert_chain(self.cert_chain_path, keyfile=self.private_key_path)
         s.load_verify_locations(capath=self.caroot_path)
         s.set_alpn_protocols(['h2'])
-        try:
-            s.set_npn_protocols(['h2'])
-        except NotImplementedError:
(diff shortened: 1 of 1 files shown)


cdecker:
The package claims support for versions 3.8 onwards, does this affect versions earlier than 3.11? If yes, we might want to adjust that support or leave this jn for now.

ErikDeSmedt:
According to this [issue](https://github.com/python/cpython/issues/90225) the functionality broke in `python3.10` which is the same release in which it was [deprecated](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_npn_protocols).

To my understanding our code only works in `python3.8` and `python3.9`.

I don't understand why we would want to keep `npn` in `gl-testing`. We only have a single client which is`gl-client` and that client does support `apln`.

cdecker:
The only reason we have the npn mode in there was that earlier versions of grpc may use it. If I remember correctly this was required to get other clients talking to the mock grpc services. Then again, you're right, we don't have any other clients really.

PR #398, opened 2024-03-27
Fix Cargo Clean

ErikDeSmedt:
To reproduce the issue do `make docker-image`, `make docker-shell`,
`cargo clean`.

The issue is that you lack the rights to delete `tmp/gltesting/target`.
The folder was added as a volume and cannot be deleted.

We keep the rust target directory in `/tmp/gltesting/target/target`
instead so it can be safely deleted.

I'd recommend to run `rm -rf /tmp/gltesting` on the host-machine after merging this update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ DOCKER_OPTIONS= \
 	--user $(shell id -u):$(shell id -g) \
 	-e TMPDIR=/tmp/gltesting/tmp \
 	-v /tmp/gltesting/tmp:/tmp/gltesting/tmp \
-	-e CARGO_TARGET_DIR=/tmp/gltesting/target \
+	-e CARGO_TARGET_DIR=/tmp/gltesting/target/target \
 	-v /tmp/gltesting/target:/tmp/gltesting/target \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -46,7 +46,7 @@ DOCKER_OPTIONS= \
 	--user $(shell id -u):$(shell id -g) \
 	-e TMPDIR=/tmp/gltesting/tmp \
 	-v /tmp/gltesting/tmp:/tmp/gltesting/tmp \
-	-e CARGO_TARGET_DIR=/tmp/gltesting/target \
+	-e CARGO_TARGET_DIR=/tmp/gltesting//target/target \
```
Randy808, line 49: extra slash before the first target

Randy808:
I was still having some permission issues so I ended up executing a shell in the container as root with:
`docker exec -u root -it <container_id> bash`

and running:
`chown -R <user> /tmp/gltesting`

ErikDeSmedt:
Is this a one time action or would you have to do it every time?

Randy808:
It was a one time action

Randy808 (approved):

Randy808 (approved):

PR #399, opened 2024-03-27
Simplified outgoing webhook methods.

Randy808:
Removed the protobuf models from the outgoing webhook method signatures in Rust.

### DIFF
--- a/libs/gl-client-py/src/scheduler.rs
+++ b/libs/gl-client-py/src/scheduler.rs
@@ -91,47 +91,31 @@ where
 
     async fn add_outgoing_webhook(
         &self,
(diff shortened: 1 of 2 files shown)


ErikDeSmedt (approved):
Looks great

nepet (approved):
I like this change, keeping as much of the api in the rust lib as possible! :+1:
We may want to merge this before #383 and resolve the conflict in #383 but it shouldn't be much more work the other way around either.

cdecker (approved):
Nice usability change 👍

ErikDeSmedt (approved):

PR #401, opened 2024-04-10
docs: Fixed syntax for code examples in getting-started.

Randy808:


ErikDeSmedt:
I haven't tested it yet. But the changes look good though

nepet:
These changes still use the `TlsConfig` instead of the `Credentials` which should not work anymore. Let me know If I can help rewrite the docs to adapt to the latest changes on main.

cdecker:
I wish we had `entangled` already, then we could just make these examples sync with files in an examples directory and test them in CI too.

Aside from that the changes look good to me :+1:

Randy808:
@nepet Changes like this are tricky because the [rust sdk on crates.io](https://crates.io/crates/gl-client/versions) and the [python release](https://pypi.org/project/gl-client/#history) (using version 0.11.1) both don't have support for the credentials change yet. Even Breez, who pins to commits, [hasn't made the move yet](https://github.com/breez/breez-sdk/blob/270016fcddf064ec9864965b18d2c95feb2ea6b3/libs/sdk-core/Cargo.toml#L19) so I feel like updating the docs to v0.2 may be premature.

Screenshot 2024-04-10 at 10 45 33 AM

nepet:
I see your point @Randy808 but it also feels weird that devs that build from the repo can not use the docs. Tagging versions may be a possible improvement here (in the future).
Do you know if mkdocs can add a "tag switcher" that allows to select a version?

How about adding the commit hash that works with the current state of the docs (for devs/users to checkout) to the docs in the meantime?

ErikDeSmedt:
We should not run ahead of ourselves on this one.

I want to prioritize consistent docs and code. Properly archiving and
tagging releases could be the second step.

cdecker:
I like the idea of a version switcher eventually. However I'd rather have the docs in sync with the code, but deploy the last released branch on `gh-pages` that gives us the right outcome:

 - We tag a commit with `latest`, that then gets deployed on `gh-pages`
 - When a user comes to the docs from some external source, they end up on a version of the docs matching the released version that downloads from crates.io
 - If the user however goes through the repo, they get a version matching the repo

I think this is the lowest amount of work for the most bang for the buck. No changes needed to the docs on the 0.1 branch, just adjusting the branch / tag we deploy to `gh-pages` from, and we keep the repo in sync with the docs.

PR #404, opened 2024-04-10
Update glcli

nepet:
This PR brings the cli back into a working state.
The following calls shouldn't fail anymore, regardless if the client uses old style separated key/cert or new style credentials:
 - `glcli getinfo`
 - `glcli scheduler upgradecreds` -> should write a `credentials.gfs` into the working dir


cdecker:
ACK 2dd615d4445ff811089b12ba9981ea5ac0c64ab4

PR #405, opened 2024-04-10
docs: Added page for webhooks.

Randy808:


docs/src/reference/webhooks.md
```diff
@@ -0,0 +1,188 @@
+## Webhooks
+Webhooks are urls that receive http requests containing event-related data. The application sending the event could be part of the hosting application or a different application entirely.
```
cdecker, line 2:
```suggestion
Webhooks are URLs that receive HTTP requests containing event-related data. The application sending the event could be part of the hosting application or a different application entirely.
```

docs/src/reference/webhooks.md
```diff
@@ -0,0 +1,188 @@
...
+	
+	webhook_uri = "https://example.com"
+	scheduler.add_outgoing_webhook(webhook_uri)
+
+	save_secret_to_db(signer.node_id(), add_webhook_response.secret);
+	```
+
+## Verifying webhook payloads
+
+Webhook payloads can be verified using the secret returned from `add_outgoing_webhook`. The secret can not be shared across nodes and is only valid for the node it was returned for. The secret serves as the key needed to validate the payload in an hmac-sha256 hash. If the payload is valid, the resulting hash should be equal to the base58-encoded value contained within the 'gl-signature' header of the request sent to the webhook.
```
cdecker, line 109: We might need to specify that the payload must be hashed and signature verified without parsing the JSON and then serializing again, as the whitespace in the JSON is not guaranteed to be reproduced faithfully. A middleware would be perfect here, but we just need to ensure the webhook server and the recipient of the webhook call agree on the cleartext down to the byte.

cdecker:
Very nice ^^

docs/src/reference/webhooks.md
Randy808: Done

cdecker (approved):
ACK https://github.com/Blockstream/greenlight/pull/405/commits/8016b39d3425c77c43449b7b2bb8386b05f18d46

PR #406, opened 2024-04-12
Additions for backend tests

nepet:
This PR introduces some helpers for the tests.


Randy808 (dismissed):

ErikDeSmedt (dismissed):

libs/gl-client/src/credentials.rs
```diff
@@ -291,8 +291,32 @@ mod model {
...
+fn load_file_or_default(varname: &str, default: &[u8]) -> Vec<u8> {
+    match std::env::var(varname) {
+        Ok(fname) => {
+            debug!("Loading file {} for envvar {}", fname, varname);
+            let f = std::fs::read(fname.clone());
+            if f.is_err() {
+                debug!(
+                    "Could not find file {} for var {}, loading from default",
+                    fname, varname
+                );
```
cdecker, line 306: Should we really allow this to go back to default almost quietly? Failing loudly here is likely cleaner.

cdecker (approved):
ACK https://github.com/Blockstream/greenlight/pull/406/commits/ddd0030bb1bbe06feac7681198ec5f589119f09a

libs/gl-client/src/credentials.rs
nepet: Would you prefer to panic or do you think raising the log level is enough?

PR #409, opened 2024-04-17
Fix shutdown signal handling in signer loop

ok300:
This PR is ported from https://github.com/breez/breez-sdk/pull/936

The original PR fixes an issue where `signer.run_once` can block for up to 10 minutes. This means that, if the signer was recently used, the signer loop would block on this call and disregard the shutdown signal for up to 10 minutes. The fix was to place the entire (inner) signer loop in a `tokio::select`, next to the shutdown signal handling.

The first part of the signer loop (scheduler initialization and `maybe_upgrade`) was extracted to a separate method as well. This part is not affected by the shutdown signal.

Note: in our version of the loop, we use `wait = false`, whereas here it is `true`. This could mean `run_once` behaves differently for you and might not block in the first place, but I'm not sure.

### DIFF
--- a/libs/gl-client/src/signer/mod.rs
+++ b/libs/gl-client/src/signer/mod.rs
@@ -18,7 +18,7 @@ use lightning_signer::bitcoin::Network;
 use lightning_signer::node::NodeServices;
 use lightning_signer::policy::filter::FilterRule;
 use lightning_signer::util::crypto_utils;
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/409/commits/61346c34a3a751ee7423f146b6e042a9886fcff1

:rocket:

PR #412, opened 2024-04-18
Added convenience methods to credentials.

Randy808:


nepet:
@Randy808 you may want to rebase this before you put any more work into it, some parts changed, e.g. We reintroduced the `fn load_file_or_default(varname: &str, default: &[u8]) -> Result<Vec<u8>>` for the `CA_CRT` and the nobody key/cert pair.

libs/gl-client/src/utils.rs
```diff
@@ -0,0 +1,31 @@
+use anyhow::{anyhow, Result};
+use crate::tls::TlsConfig;
+
+pub fn scheduler_uri() -> String {
+    std::env::var("GL_SCHEDULER_GRPC_URI")
+        .unwrap_or_else(|_| "https://scheduler.gl.blckstrm.com".to_string())
+}
+
+pub fn get_node_id_from_tls_config(tls_config: &TlsConfig) -> Result<Vec<u8>> {
```
Randy808, line 9: I moved this out to utils.rs because it was a public module. I considered putting this method in the TlsConfig, but with node_id not being directly linked to TLS certificates, I didn't think it made sense for it to be grouped in.

libs/gl-client-py/src/credentials.rs
```diff
@@ -55,6 +57,25 @@ where
             UnifiedCredentials::Device(d) => d.tls_config(),
         }
     }
+
+    fn with_identity<V>(device_cert: V, device_key: V) -> Self
```
Randy808, line 61: I made `with_identity` determine which type of credentials to return based on the presence of a node_id in the cert. This signature had to be made this way, where a specific value for the UnifiedCredentials could not be explicitly defined by the user, because the UnifiedCredentials is a TlsConfigProvider and must implement the existing trait's associated function with no value-hint in the arguments.

libs/gl-client/src/credentials.rs
```diff
@@ -177,31 +197,30 @@ impl Device {
...
 
-        // For now, upgrade is covered by recover
-        let res = scheduler
-            .recover(signer)
-            .await
-            .map_err(|e| UpgradeCredentialsError(e.to_string()))?;
-        let mut data = model::Data::try_from(&res.creds[..])
+        self.version = CRED_VERSION;
+
+        if self.rune != String::default() {
```
nepet, line 208: Should we not check on a empty rune here? I would consider using `is_empty()` for the check which is more convenient.
```suggestion
        if self.rune.is_empty() {
```

libs/gl-client/src/credentials.rs
```diff
@@ -211,12 +230,34 @@ impl Device {
     pub fn to_bytes(&self) -> Vec<u8> {
         self.to_owned().into()
     }
+
+    pub fn node_id(&self) -> anyhow::Result<Vec<u8>> {
+        get_node_id_from_tls_config(&self.tls_config())
+    }
```
nepet, line 236: Instead of deriving the `node_id` from deep within the `tls_config`, we could as well consider to add the `node_id` to the device credentials metadata (add it as a field in `Device`).

libs/gl-client/src/credentials.rs
```diff
@@ -99,6 +103,22 @@ impl TlsConfigProvider for Nobody {
...
+    {
+        let ca =
+            load_file_or_default("GL_CA_CRT", CA_RAW).expect("Could not load file from GL_CA_CRT");
+
+        Self {
+            cert: cert.into(),
+            key: key.into(),
+            ca,
+        }
+    }
```
nepet, line 121: Is it really necessary to add this to the `TlsConfigProvider` trait? I fail to see the direct concern here. Maybe an extension or a separate trait would be cleaner?

libs/gl-client/src/credentials.rs
```diff
@@ -211,12 +230,34 @@ impl Device {
...
+            load_file_or_default("GL_CA_CRT", CA_RAW).expect("Could not load file from GL_CA_CRT");
+
+        Self {
+            cert: cert.into(),
+            key: key.into(),
+            ca,
+            ..Default::default()
+        }
+    }
 }
```
nepet, line 261: Do we really need a way for `Device` credentials to be constructed from a `cert` and a `key`? Developers should load the credentials from bytes or a path as they already contain the CA and the credentials. This kind of sidesteps the idea to make this simpler and goes one step back towards the original tls_config way.

libs/gl-client/src/scheduler.rs
```diff
@@ -433,10 +434,10 @@ where
     where
         T: GrpcClient,
     {
-        let cert_node_id = self.get_node_id_from_tls_config()?;
+        let node_id = get_node_id_from_tls_config(&self.creds.tls_config())?;
```
nepet, line 437: See the other comment, we may consider to just add the node_id to the credentials metadata as this can only be called by authenticated schedulers.

nepet (commented):
Again! Nice work @Randy808

One thing that I noticed: I think we can get rid of providing the `node_id` to the `scheduler` on `scheduler::new()` altogether. We could add it to the `Device` credentials e.g. as a trait `NodeIdProvider` instead of deriving it from the `tls_config`. The non-authenticated methods `register`, `recover` could get the `node_id` from the `signer` that they get. This may allow us to completely remove `get_node_id_from_tls_config`.

What do you think?

Also: We may consider adding `with_identity` only to the `Nobody` credentials to avoid any confusion for developers about the use of the `Device` credentials that should be created from bytes or path.
I am not sure if it would make sense to change the signature of `Nobody::with` to use an `Option` for the `ca` param instead that falls back to `load_file_or_default. Let me know what you think about this.

libs/gl-client/src/credentials.rs
Randy808: This is mostly for developers who onboarded with TlsConfigs and need to upgrade their credentials.

cdecker:
I see a couple of discussions still ongoing, shall we draft this until the discussion has settled, or are these things that we should spin out?

cdecker:
Looks like this is runs into the cln-vm-manager issue that @ErikDeSmedt is addressing in #442. I will rebase once that's merged and merge this asap.

cdecker (approved):

PR #417, opened 2024-04-26
derive Serialize and Deserialize on grpc requests/responses

JssDWt:
It's basically the same PR as this one https://github.com/ElementsProject/lightning/pull/7015
But for Greenlight.
This allows easy pretty printing of responses received from the node.


ErikDeSmedt:
ACK

ErikDeSmedt (approved):

PR #442, opened 2024-05-14
CI: Do not install path-deps in `python-builder`

ErikDeSmedt:
The `python-builder` is the first stage when building the Dockerfile.
It installs all python dependencies into a `venv`.

Installing dependencies is time-consuming and we want to cache this.
Therefore, we don't want to include the path-dependencies yet. If we
would add `gl-client` and `gl-testing` we would have to rebuild the
entire docker container every time we change a line in the code-base.

In the `python-builder` we first generate a list of dependencies using
`poetry export`. We filter out the path-dependencies of this list before
installing them.

However, the script to filter out the path-dependencies is broken. We
fail to filter out the path-dependencies. This modification fixes the
problem and ensures the pipeline runs again.

### DIFF
--- a/docker/gl-testing/Dockerfile
+++ b/docker/gl-testing/Dockerfile
@@ -73,7 +73,7 @@ WORKDIR repo
 # don't have the code in the Docker-repo yet.
 RUN poetry lock
 RUN poetry export -f requirements.txt -o requirements.txt --with=dev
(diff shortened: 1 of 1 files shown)


ErikDeSmedt:
The error message given by poetry is a bit tricky. It hints to a problem with verifying the hashes.

This is an alternative to the `--without-hashes` @cdecker applied yesterday to other branches.

cdecker (approved):

PR #466, opened 2024-06-17
gl-testing: Cleaning imports

ErikDeSmedt:


ErikDeSmedt:
I've misconfigured my editor and did some clean-ups which appear to make the tests fail.

libs/gl-testing/gltesting/fixtures.py
```diff
@@ -45,7 +42,7 @@ def root_id(cert_directory):
 
 
 @pytest.fixture()
-def scheduler_id(root_id):
```
cdecker, line 48: These are needed, even if unused, because they build up the tree incrementally, and the dependency ensures the right execution order. I'm positive that is your problem. Fixtures with side-effects are nasty, but I don't know how to better do it.

PR #469, opened 2024-06-21
docs: Do not recommend installation from git

ErikDeSmedt:
Recommend using a tagged installation.


Randy808 (dismissed):

cdecker:
I think that should be 0.2 not 2.0 😋

cdecker (changes requested):
The correct version is 0.2

cdecker (approved):

PR #475, opened 2024-07-04
Trampoline client to delegate payments to the LSP

### LINKED ISSUE #470: misc: Mark the `.proto` files in `.resources` as derived
This is just because they are copies of the top-level `/proto` files, and therefore all changes would be duplicated. Marking them as hidden diff makes reviews simpler

nepet:
Adds a custom trampoline client to `gl-plugin` to enable custom trampoline payments via breez on greenlight. This covers basic functionality and basic tests.

This PR also resolves #470


cdecker (approved):
Very nice, I like the tests too, they cover quite a bit of surface :+1:

cdecker:
Now the linguist annotations work :+1:

nepet:
Rebased on `main` and switched over to `call_raw()` for `sendpay`

JssDWt:
Documenting here that there is an issue where the breez sdk gets a transport error from the trampoline_pay call when the trampoline payment succeeds. 

Other things I was wondering are:
- does this function have the same checks as the `pay` command in terms of requests waiting for the node to be ready?

libs/gl-client-py/glclient/glclient.pyi
```diff
@@ -35,7 +34,6 @@ class Credentials:
 class SignerHandle:
     def shutdown(self) -> None: ...
 
-
```
cdecker, line 38: Nit: idiomatic python requires two blank lines between top-level declarations.

libs/gl-client-py/tests/test_plugin.py
```diff
@@ -1,23 +1,28 @@
-from gltesting.fixtures import Client
+# from gltesting.fixtures import Client
```
cdecker, line 1: Delete this altogether, it is subsumed by the wildcard import below.

libs/gl-plugin/src/tramp.rs
```diff
@@ -0,0 +1,354 @@
...
+        })
+        .await
+    {
+        Ok(r) => Ok(r),
+        Err(e) => Err(anyhow!(e.to_string())),
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct SendpayRequest {
```
cdecker, line 333: I'm surprised this is necessary, shouldn't we have a Pay request like this already in cln-rpc?

nepet:
I squashed the two proto commits into one to be able to separately cherry-pick it.

libs/gl-client-py/glclient/glclient.pyi
nepet: Oh, this was done by the formater (black). Let me check if I have some ill config.

PR #483, opened 2024-07-18
signer: Capture signer rejections and report them

cdecker:
We want to capture policy signer rejections too, as they can help us identify problems for false negatives, and can be used to fine-tune policies, or act as a regression test suite should the rejection be a true positive.

Sadly we were not actually sending a rejection, due to a misplaces `?` that'd bubble up the error past the reporting spot. This PR addresses that by handling the response in a more fine-grained manner.

### DIFF
--- a/libs/gl-client/src/signer/mod.rs
+++ b/libs/gl-client/src/signer/mod.rs
@@ -316,7 +316,7 @@ impl Signer {
         }
 
         // Currently we only use a 0 unique_id and a pubkey field to allow
-        // for delegation in the future but we could also set the public 
+        // for delegation in the future but we could also set the public
         // key as the unique_id in the future and add a method that allows
         // to create new empty runes.
         let unique_id = rune.get_id();
@@ -527,7 +527,7 @@ impl Signer {
         log::trace!("State updated");
 
         // Match over root and client handler.
-        let response = match req.context {
+        let response = match req.context.clone() {
             Some(HsmRequestContext { dbid: 0, .. }) | None => {
                 // This is the main daemon talking to us.
                 root_handler.handle(msg)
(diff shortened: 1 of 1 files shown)


libs/gl-client/src/signer/mod.rs
```diff
@@ -539,15 +539,24 @@ impl Signer {
...
+                msg: format!("{:?}", e),
+                request: Some(req.clone()),
+                git_version: GITHASH.to_string(),
+            })
+            .await;
+            return Err(Error::Other(anyhow!("processing request: {e:?}")));
         }
-        .map_err(|e| Error::Other(anyhow!("processing request: {e:?}")))?;
 
+        let response = response.unwrap();
```
philipr-za, line 554: You can avoid the clone by dropping the reporter call into the existing map_err closure:
```rust
.map_err(|e| 
    {
        report::Reporter::report(crate::pb::scheduler::SignerRejection {
                msg: format!("{:?}", e),
                request: Some(req.clone()),
                git_version: GITHASH.to_string(),
            })
            .await;
          Error::Other(anyhow!("processing request: {e:?}"))
    }
)?;

```

cdecker: Clever ^^

cdecker: Yeah, sadly we can't `.await` inside of a `map_err`, but would have been nice ^^

philipr-za: Ahh, my async is rusty

philipr-za (approved):
Not massively well versed in this codebase yet but LGTM

Randy808 (approved):

PR #498, opened 2024-08-17
Revert trampoline keep-alive logic

nepet:
This reverts commit adc05479b20a920d6206df117581fa0de6db146e until we know that it is really necessary to add the custom keep-alive.


cdecker (approved):
ACK d8edce211b6f6d962131da337e9f760172840a70

PR #501, opened 2024-08-19
Add .resources dir to all crates in workspace

nepet:
In order to publish the crates individually, we need to attach our resources to them. Currently this is mostly the proto files and the tls (nobody) certificates where needed.

This PR also adds some improvements to the Makefiles (also the root Makefile) for ease of use.


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/501/commits/5725a7326ff08c79578f689b76f78cf68bb40239

PR #504, opened 2024-08-27
vls: Upgrade to the v0.12 release of VLS

cdecker:


cdecker:
@roeierez sadly, this new version does not resolve the issue I was facing earlier today, but it still addresses some issues, and we should likely just keep as much in sync with VLS as we can, so let's apply this change.

The desync issue that I'm seeing on my host (and nowhere esle for now), is a complaint about wanting to sign state number 22319 when it should be 22324. I'll dig further into how we got into this situation.

nepet (approved):
VLS confuses me a bit in a way that the `handle()` function of different handler types return different results but afaict all changes are correct.

ack 899a9baf0e4abe93c69aba266d2898ad3556220d

PR #505, opened 2024-08-28
Add awaitable channels to trampoline_pay

nepet:
This adds the `AwaitableChannels` that are already used in the internal plugin to the public plugin and adds them to `trampoline_pay`. This way we can ensure that the channel we are trying to use is actually re-established and ready to accept an htlc. Also we wait for the peer connection to re-establish before we fetch the feature bits.


cdecker (approved):
ACK 9b7a0a8e0a5641e48b012bb7ea08078e66a841d0

PR #506, opened 2024-08-28
docs: Document the Node Domain

cdecker:
We have implemented the custom error codes, but havent documented
them publicly. This fixes that :-)

### DIFF
--- a/docs/src/reference/node-domain.md
+++ b/docs/src/reference/node-domain.md
@@ -1,0 +1,142 @@
+# Node Domain
+
+The Greenlight team is committed to providing users with full control
+over their Lightning nodes, which is why the ability to offboard from
+a hosted non-custodial node to a self-hosted setup is a key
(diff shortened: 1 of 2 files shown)


docs/src/reference/node-domain.md
```diff
@@ -0,0 +1,137 @@
...
+while the node is being exported and re-deployed, the hurdle to taking
+ownership of your own node is minimal.
+
+Users who wish to run a Lightning Network node behind a NAT (Network
+Address Translation) or firewall often face significant connectivity
+issues. The primary challenge is that NATs and firewalls typically
+block incoming connections, which prevents clients (such as wallets or
+other nodes) from being able to connect to the node. This lack of
+inbound connectivity is crucial for a non-custodial hosted Lightning
+node, since
```
nepet, line 27: Seems like something is missing here.

nepet (approved):
Nice, I like that this is very detailed!

PR #511, opened 2024-09-04
Trampoline: Return payment_hash as bytes directly

nepet:
Trampoline pay returned the `payment_hash` hex encoded instead of raw bytes as other grpc calls do.


cdecker:
ACK 786a8488e91f633805b2222b98758120cb36cf3e

PR #515, opened 2024-09-23
Add rust based command line client

nepet:
This PR adds a rudimentary command line interface for gl-client written in rust. A cli in rust will ensure that we reflect changes in existing APIs onto the cli in the future.


libs/gl-cli/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+                    data_dir,
+                    network: cli.network,
+                },
+            )
+            .await?
+        }
+    })
+}
+
+//async fn getinfo(app: &App) -> Result<()> {
```
Randy808, line 100: Might be easier to keep track of what hasn't been moved yet by deleting some of these (all that's really left are listfunds and list_peer_channels)

Randy808:
I like the structure of this! 

One suggestion, that don't need to be added in this PR, is giving a heads up to the user when they don't have a signer connected. One way to do this is with a timeout for cln methods that require a signer, followed by an error message that asks the user if they have a connected signer running

libs/gl-cli/src/main.rs
nepet: Absolutely, just forgot to delete those 😅

nepet:
@Randy808 @cdecker Please let me know If you prefer passing the arguments of a node call via position or via flag. Compare `connect` or `listpays` with the other methods.

Randy808:
I prefer flags

Randy808 (approved):

libs/gl-cli/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+        if std::env::var("RUST_LOG").is_err() {
+            std::env::set_var("RUST_LOG", "debug")
+        }
+        env_logger::init();
+    }
+
+    let data_dir = match cli.data_dir {
+        Some(d) => util::DataDir(PathBuf::from_str(&d).unwrap()),
+        None => util::DataDir::default(),
+    };
```
cdecker, line 63: This is equivalent to `.map(|d| util::DataDir(PathBuf::from_str(&d)).unwrap_or_default()` which may be a bit more intuitive to read.

libs/gl-cli/src/util.rs
```diff
@@ -0,0 +1,91 @@
...
+
+pub const SEED_FILE_NAME: &str = "hsm_secret";
+pub const CREDENTIALS_FILE_NAME: &str = "credentials.gfs";
+pub const DEFAULT_GREENLIGHT_DIR: &str = "greenlight";
+
+// -- Seed section
+
+pub fn generate_seed() -> Result<Vec<u8>> {
+    let mut entropy = [0u8; 32];
+    let mut rng = rand::thread_rng();
```
cdecker, line 21: I had to quickly check whether the rng returned by this is a CSPRNG, and it seems it is: https://rust-random.github.io/rand/rand/rngs/struct.ThreadRng.html :+1:

libs/gl-cli/src/util.rs
```diff
@@ -0,0 +1,91 @@
...
+    rng.fill_bytes(&mut entropy);
+
+    let mnemonic = Mnemonic::from_entropy_in(Language::English, &entropy)?;
```
cdecker, line 24: We're not showing the mnemonic anywhere, why go through the indirection of using a mnemonic here at all?

libs/gl-cli/src/util.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+pub fn write_seed(file_path: impl AsRef<Path>, seed: impl AsRef<[u8]>) -> Result<()> {
+    let file_path = file_path.as_ref();
+    if let Some(parent) = file_path.parent() {
+        fs::create_dir_all(parent)?;
+    }
+
+    let mut file = File::create(file_path)?;
+    file.write_all(seed.as_ref())?;
```
cdecker, line 41: You might want to `file.flush()?;` as well, just to ensure the seed is written fully.

cdecker: Alternatively `file.sync_data()?` could also be useful to flush.

libs/gl-cli/src/util.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    Ok(())
+}
+
+// -- Credentials section
+
+pub fn write_credentials(file_path: impl AsRef<Path>, creds: impl AsRef<[u8]>) -> Result<()> {
+    let mut file = File::create(&file_path)?;
+    file.write_all(creds.as_ref())?;
```
cdecker, line 50: `file.flush()?;`

libs/gl-cli/src/node.rs
```diff
@@ -0,0 +1,327 @@
...
+    let mut stream = node
+        .stream_log(StreamLogRequest {})
+        .await
+        .map_err(Error::custom)?
+        .into_inner();
+
+    //TODO:  Use tokio::select!() and ctrl + c to cancel listening for log lines
+    loop {
+        tokio::select! {
+            biased;
```
cdecker, line 219: Not quite sure this is needed, since the ctrl_c signal is less likely to be ready when polled, and `biased` only applies to the cases where multiple branches are ready at the same time. It can prevent starvation of the `ctrl_c` branch though, which might have been your intention :thinking: 

Also the comment above is no longer valid.

libs/gl-cli/src/node.rs
```diff
@@ -0,0 +1,327 @@
...
+        .await
+        .map_err(Error::custom)?;
+
+    let mut node: gl_client::node::ClnClient = scheduler.node().await.map_err(Error::custom)?;
+    let res = node
+        .stop(cln::StopRequest {})
+        .await
+        .map_err(|e| Error::custom(e.message()))?
+        .into_inner();
+    println!("{:?}", res);
```
cdecker, line 260: Notice that this will always produce an error, that we suppress in the python glcli as to not disconcert users. Both hiding and showing are options, but in this case the error is desired.

libs/gl-cli/src/node.rs
```diff
@@ -0,0 +1,327 @@
...
+        .await
+        .map_err(Error::custom)?;
+
+    let mut node: gl_client::node::ClnClient = scheduler.node().await.map_err(Error::custom)?;
+    let res = node
+        .getinfo(cln::GetinfoRequest {})
+        .await
+        .map_err(|e| Error::custom(e.message()))?
+        .into_inner();
+    println!("{:?}", res);
```
cdecker, line 286: I think longer term we'll want a JSON conversion for these. Do you think it'd be possible to instrument `prost` in such a way that we generate the JSON equivalent output?

libs/gl-cli/src/node.rs
cdecker: Nevermind, comment gets removed in 9df9d47adcd33c83fc386a9716159b94aef98777

libs/gl-cli/src/main.rs
```diff
@@ -0,0 +1,303 @@
...
+mod util;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Cli {
+    /// The directory containing the seed and the credentials
+    #[arg(short, long, global = true)]
+    data_dir: Option<String>,
+    /// Bitcoin network to use
+    #[arg(short, long, default_value = "testnet", global = true, value_parser = clap::value_parser!(Network))]
```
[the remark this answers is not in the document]
cdecker: This is done in the second-to-last commit :+1:

cdecker (approved):
ACK 6a2a2b9791b0039e6fe933a5b465399519557fb1

I left a couple of comments, but no show-stoppers. Feel free to merge and we can iterate on `main` afterwards.

libs/gl-cli/src/util.rs
nepet: Flush does not seem to be needed when `File` is used as `write()` uses the OS's write already.
See the flush implementation in https://github.com/rust-lang/rust/blob/3802025f400af7817ba4874587e6a2df95abd65d/library/std/src/sys/unix/fs.rs#L884-L886

`sync_all()` on the other hand is nice to ensure that it is passed on to the fs.

libs/gl-cli/src/node.rs
nepet: That is a good question. I think this should be possible the problem here is more that `cln_rpc` and `cln_grpc` use several models (or messages) where the representation of public keys, features and other stuff is in bytes. We would need to add custom deserialization for the `Vec<u8>` fields into e.g. hex strings. Maybe the better place for this would be the upstream libs.

cdecker: I think we could likely go to cln-rpc and generate the hex conversion annotations in the generated code too 🤔

Issue #521, opened 2024-09-30
gl-client unrecoverable after network change


JssDWt:
Whenever there's a network change, like switching from wifi to mobile data, gl-client is unable to reconnect to greenlight using the same grpc channel. This means an app using gl-client would have to be killed and restarted to recover.

Related Breez SDK issue here https://github.com/breez/breez-sdk-greenlight/issues/1090

Unfortunately it seems that tonic's `Channel` doesn't handle reconnection at all: https://github.com/hyperium/tonic/issues/1254
Meaning reconnection logic should be implemented manually in every client package.

cdecker:
Thanks @JssDWt I was unaware that `tonic` does not handle reconnections well. Do you have an idea we could use to detect and remedy such a system? I guess you have more experience dealing with flaky connections on the mobile side than we do.

JssDWt:
> Do you have an idea we could use to detect and remedy such a system? I guess you have more experience dealing with flaky connections on the mobile side than we do.

We've been using golang grpc bindings in breezmobile, where reconnection is handled automatically by the grpc library. I'm looking into possible solutions.

JssDWt:
An idea is to use the timeout to detect hanging network connections. Problem is that it may be very slow, currently 10 minutes. We could use a lower timeout like this: https://github.com/hyperium/tonic/pull/662/files#diff-2dc4a5ebbcd9a8a198e55baa6958f271c1df257a9e3d6ae9c70295a4df7773deR133

And then manually reconnect whenever that timeout is hit.

cdecker:
So I think this is a fundamental tradeoff between traffic on what would otherwise be a quiescent connection and the time to discover a disconnection. It seems quite natural to me that in order to speed up recovery we'd need to increase the frequency of pings (wherever they may be implemented TCP, GRPC, or app level). That's where additional information, e.g., being told by the OS that we were asleep and we should check the connection, could help immensely: by sending a single ping after being notified by the OS that our connection may not be alive anymore would save us a lot of background pings (which also keep the mobile phone from saving battery by going to sleep).

Is there any API on iOS and Android to get such wake-up signals? And how hard would it be to integrate? We could also just mark the connection as no longer usable upon receiving the signal, keeping old connections alive just in case (can't reconnect and resume anyway), but all future calls go through newly established connections.

JssDWt:
[...]

PR #524, opened 2024-10-06
Ensure signer doesn't exit on connection change

roeierez:
We have noticed that when changing a network while greenlight (using breez sdk) is running we no longer able to create an invoice.
Digging into this I have noticed the signer exists with these logs:

```
[2024-10-06 14:26:41.256 DEBUG tonic::codec::decode:204] decoder inner stream error: Status { code: Unknown, message: "error reading a body from connection: address not available", source: Some(hyper::Error(Body, Error { kind: Io(Kind(AddrNotAvailable)) })) }
[2024-10-06 14:26:41.256 DEBUG gl_client::signer:855] Got an error from the scheduler status: Unknown, message: "error reading a body from connection: address not available", details: [], metadata: MetadataMap { headers: {} }
[2024-10-06 14:26:41.256 ERROR gl_client::signer:807] Scheduler signer loop exited unexpectedly: Err(Scheduler stream error Status { code: Unknown, message: "error reading a body from connection: address not available", source: Some(hyper::Error(Body, Error { kind: Io(Kind(AddrNotAvailable)) })) })
[2024-10-06 14:26:41.257 INFO gl_client::signer:812] Exiting the signer loop
[2024-10-06 14:26:41.257 INFO breez_sdk_core::greenlight::node_api:1220] signer exited gracefully

```
This PR ensures the signer won't exit and will indeed run forever as intended.
The same approach of `run_forever_inner` was applied to `run_forever_scheduler`

Note: while this prevents the signer from exiting and allow the app to recover there is still a gap of ~3 minutes before the new connected signer is receiving messages. Because I clearly see that after the fix the local signer reconnects successfully  I wonder if some keep alive settings on the backend side makes the scheduler "think" the old signer is still connected...

### DIFF
--- a/libs/gl-client/src/signer/mod.rs
+++ b/libs/gl-client/src/signer/mod.rs
@@ -504,7 +504,7 @@ impl Signer {
                 msg: e.to_string(),
                 request: Some(req.clone()),
                 git_version: GITHASH.to_string(),
-		node_id: self.node_id(),
+                node_id: self.node_id(),
             })
             .await;
             #[cfg(not(feature = "permissive"))]
@@ -569,7 +569,7 @@ impl Signer {
                 msg: format!("{:?}", e),
                 request: Some(req.clone()),
                 git_version: GITHASH.to_string(),
-		node_id: self.node_id(),
+                node_id: self.node_id(),
             })
             .await;
             return Err(Error::Other(anyhow!("processing request: {e:?}")));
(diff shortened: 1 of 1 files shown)


JssDWt:
Might fix https://github.com/breez/breez-sdk-greenlight/issues/1090
Might fix https://github.com/Blockstream/greenlight/issues/521

cdecker:
Nice change, thanks @roeierez :+1:

> Note: while this prevents the signer from exiting and allow the app to recover there is still a gap of ~3 minutes before the new connected signer is receiving messages. Because I clearly see that after the fix the local signer reconnects successfully I wonder if some keep alive settings on the backend side makes the scheduler "think" the old signer is still connected...

We buffer all pending requests, and will redeliver them to any signer that connects, redundantly, exactly for this case where a signer dies silently. So if the signer isn't immediately getting any pending request on reconnect we need to take a look at it. Generally speaking we let the client drive as much as possible, since it most likely has the most information on its network and connectivity state (power safe mode, network switch, etc), whereas the server side is just always on.

PR #528, opened 2024-10-11
enable trampoline for zero conf channels

JssDWt:
If the channel is unconfirmed it doesn't have a short_channel_id, but it does have a local alias. Include unconfirmed channels in the channel filter by initializing the `AwaitableChannel` with either the `short_channel_id` or the local alias. `AwaitableChannel` already works with the alias as well.


nepet (approved):
Thanks Jesse!
ACK 62628c0640f34f1fbb5c47a351de207076c16d46

PR #530, opened 2024-10-18
Fix path to rust examples in github action

nepet:
Fixes the path to the rust examples in the github action.


nepet:
Please wait before merging this, I am going to add some CI improvements as well. Currently we are blind if the docs can be built without an error on PRs. We should change that to a more robust system where we
- [ ] build docs on PR if we changed something in `docs/` or `examples`
- [ ] reuse build workflow and deploy docs on merge to main if we changed something in `docs/` or `examples`

cdecker:
Let's open a new issue for this, we need to get the docs back to working order, and what you're proposing is new functionality anyway.

PR #544, opened 2024-11-21
Make onion fields optional on lsp client hook handler

### LINKED ISSUE #541: panic potential: htlc_accepted hook may have some fields not set
The following fields might not be set in the htlc_accepted hook:
https://github.com/Blockstream/greenlight/blob/2cd4d07e5d850f8e80357540c2d90779ebdf072c/libs/gl-plugin/src/lsp.rs#L18-L23

These should all be wrapped in `Option<_>`

`unwrap` is called on deserialization here: https://github.com/JssDWt/greenlight/blob/5dc32eb0d7163c5946cd9f58e698168fd0a1652d/libs/gl-plugin/src/lsp.rs#L55

This may cause panicks deserializing the htlc_accepted payload.

nepet:
This PR makes all of the onion payload fields for the lsp clients `htlc_accepted` hook handler optional. We can not assume that all fields are set and we need to act to what we get. If the handler processes an onion payload that is missing essential fields it will stop executing and return the `continue` signal to core-lightning to pass the htlc to the next processor/handler in line.

Resolves #541

### DIFF
--- a/libs/gl-plugin/src/lsp.rs
+++ b/libs/gl-plugin/src/lsp.rs
@@ -15,8 +15,8 @@ use serde_json::Value;
 struct Onion {
     payload: tlv::SerializedTlvStream,
     short_channel_id: Option<ShortChannelId>,
-    forward_msat: Amount,
-    outgoing_cltv_value: u32,
+    forward_msat: Option<Amount>,
+    outgoing_cltv_value: Option<u32>,
     #[serde(deserialize_with = "from_hex")]
     shared_secret: Vec<u8>,
     #[serde(deserialize_with = "from_hex")]
@@ -59,7 +59,18 @@ pub async fn on_htlc_accepted(plugin: Plugin, v: Value) -> Result<Value, anyhow:
     log::debug!("Decoded {:?}", &req);
 
     let htlc_amt = req.htlc.amount_msat;
-    let onion_amt = req.onion.forward_msat;
+    let onion_amt = match req.onion.forward_msat {
+        Some(a) => a,
+        None => {
+            // An onion without an `amt_to_forward` is unorthodox and can not
+            // be processed by this plugin. Skip it.
+            return Ok(serde_json::to_value(HtlcAcceptedResponse {
+                result: "continue".to_string(),
+                ..Default::default()
+            })
+            .unwrap());
+        }
+    };
 
     let res = if htlc_amt.msat() < onion_amt.msat() {
         log::info!(
--- a/libs/gl-plugin/src/lsp.rs
+++ b/libs/gl-plugin/src/lsp.rs
@@ -161,7 +161,7 @@ where
 {
     match buffer {
         None => serializer.serialize_none(),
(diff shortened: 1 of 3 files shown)


libs/gl-plugin/src/lsp.rs
```diff
@@ -62,8 +62,12 @@ pub async fn on_htlc_accepted(plugin: Plugin, v: Value) -> Result<Value, anyhow:
     let onion_amt = match req.onion.forward_msat {
         Some(a) => a,
         None => {
-            // An onion without an `amt_to_forward` is unorthodox and can not
-            // be processed by this plugin. Skip it.
+            // An onion payload without an `amt_to_forward` is unorthodox and
+            // can not be processed by this plugin. Skip it.
+            log::debug!(
+                "lsp-plugin: got an onion payload={} without an amt forward_msat.",
```
cdecker: Add the outcome `continuing` so we don't spook people reading the logs, thinking that it is not being handled.

libs/gl-testing/tests/test_node.py
```diff
@@ -22,8 +22,7 @@ def test_node_start(scheduler, clients):
 
 
 def test_node_connect(scheduler, clients, bitcoind):
-    """Register and schedule a node, then connect to it.
```
cdecker, line 25: Feels like the changes here should have been a separate format commit. Are there any logic changes in here?

cdecker (changes requested):
All good changes, let's just reflect the `shared_secret` and `next_onion` having the correct optionality (they're not optional), and the other things are suggestions, so feel free to drop them.

libs/gl-plugin/src/lsp.rs
```diff
@@ -54,27 +54,35 @@ struct HtlcAcceptedResponse {
 const TLV_FORWARD_AMT: u64 = 2;
 const TLV_PAYMENT_SECRET: u64 = 8;
 
+/// A macro to break out of the current hook flow and return a `continue`
+/// signal to core-lightning. This is to be used when we don't know how to
+/// handle a given payload or as a shortcut in case we could identify that the
+/// incoming htlc is not part of a LSP jit channel opening.
+macro_rules! unwrap_or_continue {
```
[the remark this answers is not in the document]
nepet: This is what i considered to do first, but I fail to see how this is possible. How can we return an outer function from an inner function without calling panic (like unwrap does)? I also explored the possibility of using a [ControlFlow](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) or a custom implementation of the [Try](https://doc.rust-lang.org/std/ops/trait.Try.html) trait, using the ?-operator, but that also seemed a bit overly complicated given that we only use this in one place for now (the try trait is experimental as well).
I would suggest to leave it as is for now and If we feel that we could profit from this pattern at other places as well, we can move it upstream to the cln-plugin repo and may come up with a solution that is a bit more aligned with the chaining approach that rust follows.

nepet:
Made `shared_secret` and `next_onion` optional and rebased on `main`.

PR #547, opened 2024-11-27
docker: Updated Ubuntu from 20.04 to 22.04

ShahanaFarooqui:
gltestserver requires GLIBC v2.34 or higher, but Ubuntu 20.04 ships with v2.31.


cdecker:
For some reason this PR is not triggering the CI run, and we can't merge it as is.

PR #548, opened 2024-11-29
client: Tweak the keepalive interval and timeout

### LINKED ISSUE #428: Connection is getting dropped with `transport error`
This is a tracking issue for the `transport error` issue we are seeing since we activated the GL-LB for client -> node connections.

The root cause is a short loadbalancer keepalive timeout of approximately 30 seconds, in combination with calls that take longer than 30 seconds, such as some `pay` calls. The loadbalancer drops the connection after the connection has been idle, i.e., not transferring any data. Usually the client connection is configured to send keepalive messages (`PING`) when the connection is idle, while a call is pending (also referred to by the generic term `stream`, as everything in grpc is a stream). The configuration of the connection is here:

https://github.com/Blockstream/greenlight/blob/86c43646b0070b984f54bebfe245dd85c363cc44/libs/gl-client/src/node/mod.rs#L106-L112

And it configures keepalives to be very aggressive, and also enabled if there is no call pending. However, running with `RUST_LOG=trace` we can see that no pings are effectively being sent over the wire. This is strange as the signer is configured the same way, but that one is sending pings (it also has an active stream to `stream_hsm_requests` so that might be a difference).

The documentation on the semantics of keepalives is here: https://grpc.io/docs/guides/keepalive/

cdecker:
As reported by the Breez team the cause for the `transport error` when
attempting a `trampoline_pay` appears to be that even short
interruptions in connectivity kill the TCP connections. The long 90s
timeout is causing `gl-client` to patiently sit there for the 90s,
ignoring all errors in the hope of getting the TCP connection
back. When it then attempts to re-establish the connection, any
pending call gets the dreaded `transport error`.

Reducing this to 5 seconds timeout and a 1s interval should minimize
the chances of issuing a command during the broken phase.

Suggested-by: Roei Erez <@roeierez>
Suggested-by: Jesse De Wit <@JssDWt>
Closes #428

### DIFF
--- a/libs/gl-client/src/lib.rs
+++ b/libs/gl-client/src/lib.rs
@@ -70,7 +70,7 @@ pub use lightning_signer::bitcoin;
 pub use lightning_signer::lightning;
 pub use lightning_signer::lightning_invoice;
 
-pub(crate) const TCP_KEEPALIVE: Duration = Duration::from_secs(5);
-pub(crate) const TCP_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(90);
+pub(crate) const TCP_KEEPALIVE: Duration = Duration::from_secs(1);
+pub(crate) const TCP_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(5);
 
 pub mod runes;


libs/gl-client/src/lib.rs
```diff
@@ -70,7 +70,7 @@ pub use lightning_signer::bitcoin;
 pub use lightning_signer::lightning;
 pub use lightning_signer::lightning_invoice;
 
-pub(crate) const TCP_KEEPALIVE: Duration = Duration::from_secs(5);
-pub(crate) const TCP_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(90);
+pub(crate) const TCP_KEEPALIVE: Duration = Duration::from_secs(1);
+pub(crate) const TCP_KEEPALIVE_TIMEOUT: Duration = Duration::from_secs(5);
```
roeierez, line 74: I think it makes sense to set the keep alive timeout not more than the interval  otherwise we may end up sending a second keepalive packet before the first one is  ack'd.
Any reason you see not to set it for 1 second?

cdecker: It's a balance of eager timeout vs keeping connections more stable. The timeout is counted for each packet, while the interval tells us how to send the ping, the timeout tells us how long to wait for pings. Here we pipeline up to 5 pings before we require the first pong to be back. So technically, if we send 2 pings, the first one gets replied to, the second one doesn't, then at second 6 we reconnect. If we set the interval and timeout to the same, we'd wait for 10 seconds, because we send only every 5s, and wait for a reply for 5s.

Pipelining pings can help reduce the time to recovery, without increasing the sensitivity to latency. Timeout on the other hand controls latency sensitivity, and as mentioned, I'm hoping that for interruptions <5s the TCP connection can just resume, delivering a late pong but keeping it inside of the timeout. Overall I think this will minimize the number of failed grpc calls.

roeierez: > If we set the interval and timeout to the same, we'd wait for 10 seconds, because we send only every 5s, and wait for a reply for 5s.

I actually commented about making the timeout the same as the interval (both 1 seconds) which means sending every 1 seconds and waiting up to 1 second for ack, lowering the max time to detect network change to 1 second overall and lowering even more the chances to get transport errors (if the call falls into the timeout time frame).
I guess the main question is do we foresee that on a stable connection we will have more than 1 second latency for replying a single tcp packet?
I agree that this fix by itself is a significant improvement.

roeierez: In my tests tracing the keep alive packets I see response in milliseconds but perhaps my bandwidth is better than the average and we should prepare for higher latency. I am ok to start with 5 seconds and gather some feedback from the field.

PR #549, opened 2024-11-29
testserver: Add a couple of usability improvements

cdecker:
The `.envrc` file can make it simpler to configure your shell. In
addition we add `Makefile` targets.


cdecker:
There is a breaking change in `httpx`: https://github.com/encode/httpx/pull/3022

I'll fix this one and the other two PRS up asap

PR #550, opened 2024-12-03
tramp: never create zero value htlcs

JssDWt:
If a spendable channel balance exactly equals the amount, trampoline payment would create a zero value htlc. Never do that!


nepet (approved):
ack 2aa36656a8b9f88dcacfe3540b749afdc59874a2

PR #572, opened 2025-02-05
Remove explicit clnvm run

ShahanaFarooqui:
- Removed `RUN uv run clnvm get-all` since it fails with `No such file or directory (os error 2)` and is redundant, as the command runs later with `gltestserver run`.

- Set `GL_TESTING_IGNORE_HASH` to False to suppress the warning: "Checking the hash of remote versions is disabled, which is unsafe. Try to unset GL_TESTING_IGNORE_HASH."


cdecker:
ACK https://github.com/Blockstream/greenlight/pull/572/commits/67c3fdd5ef6b598e9dd64c6568b216d5b9029025

PR #576, opened 2025-02-14
Updated javascript example

ShahanaFarooqui:
Fixed bug due to encodeDelimited function.


cdecker:
ACK b79af426f7b0dd64743c2c77151022e1efad3422

cdecker:
I think you might be mixing `grpc-status` defined [here](https://grpc.io/docs/guides/status-codes/), and HTTP status codes. Can you ensure that you are using the right number range for the header `grpc-status`?

cdecker:
Perfect, thanks for checking, that puts my mind at ease :-)
Merging this as soon as I get CI to work properly :sweat_smile:

PR #579, opened 2025-03-03
Remove Legacy Lightning-RPC Client

nepet:
The plugin still had some references to a custom lightning-rpc client here and there which made it dificult to keep the `greenlight.proto` in sync. This PR removes this rpc client and syncs the `greenlight.proto` for the plugin and the signerproxy.


libs/gl-plugin/src/node/mod.rs
```diff
@@ -114,10 +106,11 @@ impl PluginNodeServer {
             use tokio::time::{sleep, Duration};
 
             // Move the lock into the closure so we can release it later.
-            let rpc = rrpc.lock().await;
+            let mut rpc = cln_rpc::ClnRpc::new(rpc_path.clone()).await.unwrap();
```
cdecker, line 109: We got a bit unlucky here: we use the `Mutex` to prevent incoming JSON-RPC calls, before the JSON-RPC socket is ready. This is done by acquiring the lock during startup and releasing it once the socket file is present. The internal use-cases are ok, since they by definition are invoked only after the node started and the socket is ready.

The change below however breaks this, by not locking in `get_rpc` it is now possible for incoming grpc calls to be directly forwarded (not waiting behind a lock) and so they will fail. We will most likely notice this as the startup commands (i.e., the ones that caused the node to start) will fail in a large number of cases.

Please leave the `Mutex` in `get_rpc`

libs/gl-plugin/src/node/mod.rs
```diff
@@ -475,10 +464,16 @@ impl Node for PluginNodeServer {
     ) -> Result<Response<pb::Empty>, Status> {
         self.limit().await;
         let gl_config = req.into_inner();
-        let rpc = self.get_rpc().await;
+        let mut rpc = self.get_rpc().await.map_err(|e| {
+            tonic::Status::new(
+                tonic::Code::Internal,
```
cdecker, line 469: And this is the error messages all "early calls" will run into. Not the best UX.

libs/gl-plugin/src/pb.rs
```diff
@@ -1,336 +1,13 @@
 tonic::include_proto!("greenlight");
-use self::amount::Unit;
```
cdecker, line 2: This deletion has been so long in the making, thank you ^^

cdecker (changes requested):
Just one regression: a lock is required to ensure that the node has its JSON-RPC socket available before we can try to talk to it.

libs/gl-plugin/src/node/mod.rs
nepet: I moved the rpc client into a `Mutex` inside of a `OnceCell` for a clean  one-time initialization.

libs/gl-plugin/src/node/mod.rs
```diff
@@ -29,6 +30,32 @@ pub use wrapper::WrappedNodeServer;
 static LIMITER: OnceCell<RateLimiter<NotKeyed, InMemoryState, MonotonicClock>> =
     OnceCell::const_new();
 
+static RPC_CLIENT: OnceCell<Arc<Mutex<cln_rpc::ClnRpc>>> = OnceCell::const_new();
+
+pub async fn get_rpc<P: AsRef<Path>>(path: P) -> Arc<Mutex<cln_rpc::ClnRpc>> {
```
cdecker, line 35: Looking good, but I wonder if `OnceCell` is already sufficient to get the semantics we want: the `init` function is guaranteed to be called at most once. This means OnceCell is already enforcing locking behavior. So as long as we can ensure that the first `init` call returns only once the socket exists, that is already sufficient, and we do not need to nest the `Mutex` inside the `OnceCell`.

Let's discuss that during our next meeting, merging this one in the meantime.

cdecker (approved):

PR #585, opened 2025-03-06
Check for `spendable_msat` on trampoline pay

nepet:
We had some incidents where a trampoline payment created `0msat` which lead to failed payments that actually did not fail. In order to avoid this we filter out channels that have a `spendable_msat` value of `0` or below the `minimum_htlc_out_msat`.

### DIFF
--- a/libs/gl-plugin/src/tramp.rs
+++ b/libs/gl-plugin/src/tramp.rs
@@ -32,6 +32,8 @@ const PAY_UNPARSEABLE_ONION_MSG: &str = "Malformed error reply";
 const PAY_UNPARSEABLE_ONION_CODE: i32 = 202;
 // How long do we wait for channels to re-establish?
 const AWAIT_CHANNELS_TIMEOUT_SEC: u64 = 20;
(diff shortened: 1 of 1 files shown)


libs/gl-plugin/src/tramp.rs
```diff
@@ -347,6 +339,45 @@ pub async fn trampolinepay(
...
+            continue;
+        }
+
+        if (channel.spendable_msat + acc) <= amount_msat {
+            choosen.push((channel.short_channel_id, channel.spendable_msat));
+            acc += channel.spendable_msat;
+        } else {
+            let rest = amount_msat - acc;
+            choosen.push((channel.short_channel_id, rest));
+            acc += rest;
```
JssDWt, line 369: There's still a case where `rest` does not add up to the minimum htlc amount. Do you know a good way to handle that?

nepet: Right, we may want to check and skip those channels as well. :thinking: 
I'll add a follow up.

PR #586, opened 2025-03-06
Trampoline: Improve channel selection logic

nepet:
The way we allocated channels for a trampoline payment could lead to a case where we could get stuck selecting channels for a payment when the rest amount was lower than the lower bound of the channel.

This PR introduces a new selection logic that tries to be greedy first but will split payments more carefully if it is actually needded.

Reviewers: Please check carefully and let me know if you need more tests.

### DIFF
--- a/libs/gl-plugin/src/tramp.rs
+++ b/libs/gl-plugin/src/tramp.rs
@@ -232,6 +232,8 @@ pub async fn trampolinepay(
                 min_htlc_out_msat,
             });
         })
+        .filter(|ch| ch.spendable_msat > 0)
+        .filter(|ch| ch.spendable_msat > ch.min_htlc_out_msat)
         .collect();
 
     channels.sort_by(|a, b| b.spendable_msat.cmp(&a.spendable_msat));
--- a/libs/gl-plugin/src/tramp.rs
+++ b/libs/gl-plugin/src/tramp.rs
@@ -192,7 +192,7 @@ pub async fn trampolinepay(
 
     debug!("overpay={}, total_amt={}", overpay, amount_msat);
 
-    let mut channels: Vec<ChannelData> = rpc
+    let mut channels: Vec<Channel> = rpc
         .call_typed(&cln_rpc::model::requests::ListpeerchannelsRequest { id: Some(node_id) })
         .await?
         .channels
@@ -226,7 +226,7 @@ pub async fn trampolinepay(
                     return None;
                 }
             };
-            return Some(ChannelData {
+            return Some(Channel {
                 short_channel_id,
                 spendable_msat,
                 min_htlc_out_msat,
@@ -448,11 +448,11 @@ async fn do_pay(
 }
 
 async fn reestablished_channels(
-    channels: Vec<ChannelData>,
+    channels: Vec<Channel>,
     node_id: PublicKey,
     rpc_path: PathBuf,
     deadline: Instant,
-) -> Result<Vec<ChannelData>> {
+) -> Result<Vec<Channel>> {
     // Wait for channels to re-establish.
     crate::awaitables::assert_send(AwaitableChannel::new(
         node_id,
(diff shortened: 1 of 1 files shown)


libs/gl-plugin/src/tramp.rs
```diff
@@ -488,15 +454,99 @@ async fn reestablished_channels(
...
+        }
+
+        // Base case: We have processed all channels and could not find a
+        // solution.
+        if idx >= channels.len() {
+            return;
+        }
+
+        // Option 1: Skip the current channel.
+        rec(channels, idx + 1, target_msat, current, best);
```
JssDWt, line 511: This starts by skipping the largest channels. I think it would be best to start with the ideal case, and the most probable case: the htlc fits in a single channel (the first channel). Basically the base cases that would work most of the time is 'take the maximum from each channel until the amount fits. 

I think the algorithm here should return immediately once 'a' solution has been found. It does not have to be the best solution. 

So skipping the current channel should happen last. (unless channels are ordered from smallest to largest).

libs/gl-plugin/src/tramp.rs
```diff
@@ -488,15 +454,99 @@ async fn reestablished_channels(
...
+
+        // If the remaining target amount is below the channel's lower bound we
+        // can not use it.
+        if target_msat < lower {
+            return;
+        }
+
+        // Try allocations from the upper bound down to the lower bound to
+        // maximize channel usage.
+        for alloc in (lower..=upper).rev() {
```
JssDWt, line 529: This will iterate a lot? min_msat being 1 and amount being 1000 sat (1M msat) will iterate 1M times per channel? So 1M * 1M for 2 channels?

What if (thinking out loud here) you would start with the maximum per channel. Then if that doesn't work, try to leave the minimum minimum (so the lowest min_htlc_out_msat of the remaining channels), if that doesn't work leave the second lowest minimum, etc. So the allocation of the current channel would become min(spendable_msat, target_msat - min_htlc_out_msat of the next channel).

libs/gl-plugin/src/tramp.rs
nepet: I was assuming that it might be helpful to fully drain channels to allow for re-allocation or consolidation of those channels, and since we sort the channels: largest spendable msat first, What should happen here is that we prefer to drain smaller channels fist. Choosing the best case instead of just choosing a case that works was part of the same assumtion I made. I can add a test to verify this behavior or we can change to any other policy you prefer.

libs/gl-plugin/src/tramp.rs
nepet: Yes that will iterate a lot and I like your improvement! What if we would, instead of choosing the lowest `min_htlc_out_msat`, choose the `min_htlc_out_msat` of the second smallest channel to try to drain as many channels as possible to  make them available for re-allocation?

libs/gl-plugin/src/tramp.rs
JssDWt: That does sound good actually, to drain the smaller channels first.

libs/gl-plugin/src/tramp.rs
JssDWt: That sounds ok. Do add a fallback to the third smallest channel, then the fourth smallest, etc, if it doesn't work.

nepet: Sure, we want to recursively try bigger channels until we run out of options.

nepet:
@JssDWt If you don't mind it would be nice to get your oppinion on the improved version!

libs/gl-plugin/src/tramp.rs
```diff
@@ -488,15 +468,201 @@ async fn reestablished_channels(
...
+
+        // Channel is drained or unusable, skip it.
+        if ch.spendable_msat < ch.min_htlc_out_msat || ch.spendable_msat == 0 {
+            return try_allocate(&channels[1..], target_msat, allocations);
+        }
+
+        // Each channel has an upper and a lower bound defined by the minimum
+        // HTLC amount and the spendable amount.
+        let lower = ch.min_htlc_out_msat;
+        let upper = ch.spendable_msat.min(target_msat);
```
JssDWt, line 600: Let's add a check that `upper` is not below `lower`.

JssDWt (commented):
It looks very good to me! I can't find any real issues on the first pass.

cdecker:
Thanks guys, apart from the check that @JssDWt proposes this looks great, and we can add that safety check in a follow up. @nepet are you ok with me merging this as is?

PR #588, opened 2025-04-02
clnvm: Add v24.11gl1 and make it a bit more verbose

cdecker:
The `clnvm` tool was way too quiet for my liking, so I added logging to `stderr` not to interfere with tooling that may already rely on the tool printing the paths of versions to `stdout`.

I then also added the new version to deploy in `prod`.


nepet (approved):
Looks good to me!

PR #589, opened 2025-04-04
signer: Add CLN v24.11 to supported versions

cdecker:
This PR adds support for CLN v24.11. It'll trigger an upgrade on the
node as soon as it connects the first time, and will act as its signer
from thereon.

We're keeping it a draft, while we verify the core functionality, and
will merge as soon as it is deemed stable. Feel free to use this
signer version, however be aware that there is currently no way to
downgrade once upgraded, so use test nodes.


cdecker:
The new version has been tested extensively both internally and externally, so we're merging this as is :+1: Thanks everyone for testing.

PR #590, opened 2025-04-08
Upgrade `cln_grpc` and `cln_rpc`

nepet:
In order to keep up with changes in `tonic` and `prost` as well as changes in the upstream cln repo we need to upgrade those crates once in a while.

The newly added streams from `cln_grpc` are left to be implemented in the future as we need to access a module that is currently private.


libs/gl-plugin/src/awaitables.rs
```diff
@@ -324,28 +320,10 @@ async fn billboard(
...
-            .into_iter()
-            .nth(0)
-            .ok_or(Error::Channel("Has no channels list"))?
-            .into_iter()
-            .filter(|c| c.short_channel_id == Some(scid))
-            .nth(0)
-            .ok_or(Error::Channel("No channel with scid"))?
-            .status
-            .ok_or(Error::Channel("No amount found"))?)
+        return Err(Error::Service(format!(
```
cdecker, line 323: Ok, this means we will deprecate any pre-v23.05 node in the `prod` system before rollout. That is intended I assume?

cdecker: For reference, that is currently about <1% of nodes, and <0.01% of active nodes.

cdecker (approved):

PR #593, opened 2025-04-13
gl-testing: enable multi-arch protoc download for Apple Silicon

YusukeShimizu:
Enable multi-architecture support to ensure proper protoc downloads based on the build environment.
This fixes issues on Apple Silicon (ARM64) where the x86_64 binary was previously being downloaded,
leading to the missing ld-linux-x86-64.so.2 error.

* Use the same protocol buffer compiler version across architectures
* Ensure Docker build success on Apple Silicon

### DIFF
--- a/docker/gl-testing/Dockerfile
+++ b/docker/gl-testing/Dockerfile
@@ -109,10 +109,18 @@ RUN wget -q https://github.com/cloudflare/cfssl/releases/download/v1.5.0/cfssljs
 # STAGE: protoc-downloader
 # ----------------------------------------
 FROM ubuntu:22.04 AS protoc-downloader
+ARG TARGETPLATFORM
+
 RUN apt update && apt install -qqy wget unzip
-RUN mkdir /tmp/protoc && \
+RUN PLATFORM=$( \
+    case ${TARGETPLATFORM} in \
+    linux/amd64 ) echo "x86_64";; \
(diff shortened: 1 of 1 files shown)


cdecker:
Very good contribution. I am surprised that the arch is passed through to the docker emulation directly, and not using `qemu` with x86 or x86_64. Today I learned :-)

If you undraft I will merge this asap ^^

YusukeShimizu:
Thank you for confirming. The current support is still insufficient.
To support building on ARM64, not only `protoc` but also all dependent binaries—such as bitcoin-core, cfssl, and cln—must be similarly addressed. In particular, cln (clm-vm) does not currently provide an ARM64 binary, raising concerns about increased support and maintenance costs.

Abandoning this approach and instead updating the `make` command to force emulation by specifying `--platform linux/amd64` at build time is a simpler solution. However, running an Intel-based container on an Arm-based machine remains a best-effort measure, with associated risks of performance degradation and potential bugs.

I would appreciate your feedback.

cdecker:
Indeed, we have been developing exclusively on `amd64`, and adding other archs, distros, and OSs may be a lot of surface. Parametrizing the binaries being downloaded is an excellent first step, and may be sufficient. Let's see if it works out :-)

YusukeShimizu:
I removed the draft status from this PR.

cdecker (approved):

Issue #563, opened 2025-01-16
gltestserver needs a diet


cdecker:
The recently introduced gl-testserver is run from a docker image that
contains all the necessary dependencies, and some more. The image
weighs in at over 9GiB, which even developing locally is a bit
much. It means that the developer disk is pointlessly being filled,
but also that a build and download of the docker image is a lengthy
process.

Putting it on a diet to weigh less, and come with fewer optional
dependencies is likely a good idea. These are the common threads that
we might want to investigate:

 - [ ] Rust compilation artifacts should be kept in a builder image, and
   selectively copied over.

PR #598, opened 2025-04-28
testserver: slim testserver image

YusukeShimizu:
Issue: [#563](https://github.com/Blockstream/greenlight/issues/563)

Reduces the image size for the gl-testserver by switching to a multi-stage Dockerfile build.  
The previous image was around 9 GiB, causing long build/fetch times and taking up unnecessary space in development environments. With this change, the image footprint has been reduced to approximately 500MB.

```sh
docker images
REPOSITORY                      TAG       IMAGE ID       CREATED             SIZE
gltestserver                    latest    9fe837b67bad   About an hour ago   525MB
```

Also enable multi-architecture support.

### DIFF
--- a/docker/gl-testserver/Dockerfile
+++ b/docker/gl-testserver/Dockerfile
@@ -1,84 +1,187 @@
-FROM ubuntu:22.04 AS python-builder
+############################
+# Stage 1: builder
+############################
+FROM rust:slim-bullseye AS builder
 
+#---------------------------------------------------------------------
+# Build-time arguments & environment variables
+#---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


YusukeShimizu:
[The Check Self / Linux (pull_request) CI test](https://github.com/Blockstream/greenlight/actions/runs/14698348884/job/41243403242?pr=598) failed, so I’m looking into what went wrong. It doesn’t seem to be caused by the PR itself, so it’s possible that the test might be flaky.

cdecker:
This is great work, thank you very much @YusukeShimizu :pray:

Issue #597, opened 2025-04-25
glcli: Large responses are rejected


cdecker:
The new rust-based `glcli` appears to be enforcing the 4MB response size limit:

```
Error, message length too large: found 4362728 bytes, the limit is: 4194304 bytes
```

@nepet: if I remember correctly there was a parameter we could pass to the stub to configure that limit. Since we trust the server when it comes to responses, we can lift that limit, to 32MB for example.

YusukeShimizu:
Would it be okay for me to take care of this issue? We can adjust the [decoding message size limit](https://docs.rs/tonic/latest/tonic/client/struct.Grpc.html#method.max_decoding_message_size).  
It defaults to 4MB, so it looks like the default setting is currently in effect.

cdecker:
Indeed, we have encountered this in the past, when syncing the gossip internally. We could just lift the maximum size the client accepts, since we have control over the server-side, and don't need to worry about DoS (a client asking for too large a response is just hammering their own node, not affecting others). And the client is likely in the best position to judge whether it can cache and process large responses, and should set limits accordingly.

nepet:
I think this is more an issue on `gl-client` as `glcli` uses `scheduler.node()` to get a grpc client from `gl-client`. As @YusukeShimizu proposed this should be fixable by setting a bigger size limit when constructing the grpc clients.

If you want to have a take on this @YusukeShimizu, you can find them here:
https://github.com/Blockstream/greenlight/blob/ed1f18b4a965e92ff7e38259ef2f03f011a02a8e/libs/gl-client/src/node/mod.rs#L39-L55

We may want to consider to make this a configurable option (maybe via something like `scheduler.node(max_size) or an env var` with a sane default.

cdecker:
Addressed by https://github.com/Blockstream/greenlight/pull/599

PR #599, opened 2025-05-07
glclient: lift default decoding size

YusukeShimizu:
https://github.com/Blockstream/greenlight/issues/597 
Lift the maximum decoding message size from 4MB to 32MB by default.

Prevent large gRPC responses from failing.
Make it configurable via the Node builder.
We trust the server logic, so increasing the default is safe.


cdecker (approved):

cdecker:
Very nice, thank you @YusukeShimizu 🤗

PR #625, opened 2025-10-25
gltestserver: Fix Dockerfile build failures and dockerignore .venv

ShahanaFarooqui:
- Locked Python dependency to 3.10 because Python 3.9's end-of-life is 31st October, 2025
- Updated base images versions
- Removed unnecessary installations
- Simplified command instruction using uv run
- Fixed runtime errors


cdecker:
Rebased on top of #631 now that it's merged.

PR #632, opened 2025-11-11
feat(node): Add `lsps2-invoice` support to the plugin

cdecker:
This PR adds rudimentary support for the LSPS2 protocol, based on @nepet's CLN https://github.com/ElementsProject/lightning/pull/8569 PR adding these features.

To make this work we had to do a bit of yak shaving:

 - The CLN builder for the GL versions of the node software now also builds a `main` version, which uses a snapshot of `master` from CLN. This version is used as a floating target for the network to use, when testing.
 - The CLN builder also remixes the CLN versions, adding newer plugins that are backwards compatible to the v25.05gl1 version, so it can get lsps support and a more modern `renepay`. We will use this in future to ensure even old versions get the upside of updates on the development branch.
 - We now pull a main version as well
 - We needed to enable the lsps-service
 - We needed an lsps-policy plugin, so queries return something

Notice that the test I added so far only tests the lsps-invoice call, not the actual execution. @nepet I think this is likely easiest for you to add and test.

### DIFF
--- a/libs/gl-client-py/tests/test_plugin.py
+++ b/libs/gl-client-py/tests/test_plugin.py
@@ -234,4 +234,4 @@ def test_lsps_plugin_calls(clients, bitcoind, node_factory, lsps_server):
     # Only one routehint, with only one hop, the LSP to the destination
     assert len(inv.route_hints.route_hints) == 1
     rh = inv.route_hints.route_hints[0]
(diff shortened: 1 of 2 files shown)


cdecker:
Rebased on top of #631 now that it's merged.

PR #697, opened 2026-04-09
bindings: update build for swift bindings

lvaccaro:
1. Add support for legacy x86_64 ios simulator and merge ios-sim libraries into one with `lipo`
2. Apple compliant framework distribution:
- creating swift bindings
- creating a wrapper project with `xcodegen`, instead maintain a full .xcproject
- archiving both platforms
- assembling the final xcframework


cdecker:
Thank youuuu 🙏

PR #700, opened 2026-04-14
glcli: fundchannel rpc make new channels private

Lagrang3:
Make new channels private by default.


cdecker:
Excellent change, thanks. We migth want to revisit for the enterprise variant as they'll want to make it routable, but for end-users this is the right thing to do.

PR #701, opened 2026-04-21
gl-sdk: Add LNURL pay, withdraw, and address resolution

cdecker:
## Summary

Adds LNURL support to `gl-sdk` (and the underlying `gl-client`), exposed through UniFFI and NAPI bindings. The change is purely additive — no existing APIs were removed or had their signatures changed.

### New top-level types (11 classes)
- `ResolvedLnUrl` — result of resolving a bare LNURL / Lightning Address
- `LnUrlPayRequest`, `LnUrlPayRequestData`, `LnUrlPayResult`, `LnUrlPaySuccessData`
- `LnUrlWithdrawRequest`, `LnUrlWithdrawRequestData`, `LnUrlWithdrawResult`, `LnUrlWithdrawSuccessData`
- `LnUrlErrorData`, `SuccessActionProcessed`

### New methods on `Node`
- `resolve_lnurl(input: str) -> ResolvedLnUrl`
- `lnurl_pay(request: LnUrlPayRequest) -> LnUrlPayResult`
- `lnurl_withdraw(request: LnUrlWithdrawRequest) -> LnUrlWithdrawResult`

### `InputType` extensions
`parse_input()` now also recognizes bare LNURLs and Lightning Addresses; `InputType` gained `is_ln_url()` / `is_ln_url_address()` variant checks alongside the existing `bolt11` / `node_id` ones.

### Supporting work
- `gl-client`: LNURL protocol primitives covering LUD-01/03/06/09/10/16
- `gl-sdk-napi`: matching TypeScript/Node bindings for the new types and methods
- `gl-testing`: CLN-backed LNURL server fixture plus integration tests for pay/withdraw flows
- Refactor: `fetch_invoice()` and `build_withdraw_callback_url()` extracted as free functions in `gl-client` for reuse across callers

## Test plan
- [x] `task sdk:test` passes (UniFFI Python bindings + integration tests)
- [x] `cargo test -p gl-client` passes (LNURL protocol unit tests)
- [x] `cargo test -p gl-sdk` passes
- [x] NAPI bindings build and tests pass (`libs/gl-sdk-napi`)
- [x] `task testing:check` passes (LNURL server fixture + `test_lnurl.py`, `test_lnurl_server.py`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #702, opened 2026-04-23
gl-client/gl-sdk: Harden LNURL-pay against real-world services

angelix:
Gaps surfaced when paying stacker.news exposed several mismatches between the current implementation and what wallets need in practice. Changes align gl-sdk's LNURL-pay surface with the LUD specs and with what Breez SDK does for the convention-level gaps:

- Parse LUD-06 service errors: callback responses with {"status":"ERROR"} are now recognised and surfaced as LnUrlPayResult::EndpointError with the service's reason, instead of failing JSON deserialization.
- Add LnUrlPayResult::PayError { payment_hash, reason } so CLN pay-side failures return structured results rather than Error::Rpc.
- Pre-flight amount/comment validation in gl-sdk::lnurl_pay to reject out-of-bounds requests before any network round-trip.
- Drop the description_hash == SHA256(metadata) check that rejected compliant-enough services whose metadata embeds per-request data.
- Skip the empty `comment` query param in the callback URL when the caller passes None or an empty string.
- Enforce LUD-09/10 bounds on SuccessAction payloads (Message/Url description ≤ 144, AES description ≤ 144, ciphertext ≤ 4096, IV exactly 24 chars) before any AES decryption.
- Validate the invoice's BOLT-11 currency prefix against the node's configured network; thread `network` through Node::with_signer so the check has something to compare against.
- Validate that a URL success action's domain matches the callback domain, with an opt-out via LnUrlPayRequest.validate_success_action_url (defaults to true).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1647,6 +1647,7 @@ dependencies = [
  "tonic 0.11.0",
  "tracing",
  "uniffi",
+ "url",
 ]
 
 [[package]]
--- a/libs/gl-client/src/lnurl/models.rs
+++ b/libs/gl-client/src/lnurl/models.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, ensure, Result};
 use async_trait::async_trait;
 use log::debug;
 use mockall::automock;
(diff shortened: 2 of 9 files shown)


cdecker:
Thanks @angelix, I'm surprised by the fact that the metadata match check is giving issues, then again, why am I surprised that the spec is not being followed? :wink: 
There is just a missing field in an initializer, and two match statements that are not exhaustive. Should be good to go as soon as it compiles.

cdecker:
Fixed the error in #701 and rebased on top of it. Merging as soon as green.

PR #704, opened 2026-04-24
feat(gl-sdk): NodeState snapshot with UTXO-safe balance accounting

angelix:
Unified node_state() snapshot aggregating get_info, list_peer_channels, and list_funds into a single NodeState record with balances, capacity, utxos, and connected-channel-peer set.

Balance accounting fixes vs. a naive implementation:

- Reserved UTXOs (in-flight PSBTs) are excluded from on-chain balance. FundOutput now carries a `reserved` field sourced from CLN's listfunds.outputs[].reserved; node_state() skips reserved outputs.
- Closing-channel double-count fixed. Onchain-state channels only contribute to pending_onchain_balance_msat when we initiated the close and our payout is still timelocked (DELAYED_OUTPUT_TO_US), otherwise the payout is already visible as a wallet UTXO and would be counted twice. PeerChannel now exposes `closer` (ChannelSide) and `status` to support this gate.
- ChannelState::from_i32 unknown fallback now maps to a new ChannelState::Unknown variant rather than Onchain, which silently counted unmapped states as closing. Unknown is treated as neither open nor closing by balance math.
- Immature outputs (confirmed but timelocked, e.g. coinbase maturation) now surface as their own immature_onchain_balance_msat instead of being silently dropped. OutputStatus match is exhaustive — future variants become a compile error rather than silent zero.
- The three RPCs run concurrently via tokio::join! rather than serially, cutting node_state() latency ~3x on remote nodes.

Mobile-developer API surface:

- id, color, channel_id, peer_id, funding_txid, txid fields use lowercase hex Strings instead of raw Vec<u8>, eliminating the encode-for-display step at every call site.
- New aggregate fields: total_onchain_msat (confirmed + unconfirmed + immature), total_balance_msat (everything the user owns), spendable_balance_msat (send-screen gate). All msat for consistency.
- max_chan_reserve_msat (protocol reserve locked across channels).
- utxos: Vec<FundOutput> for coin-control UIs. Excludes spent.
- Docstrings on channels_balance_msat / max_payable_msat explicitly name their roles: home-screen display vs send-button gate.
- Docstrings on num_active/pending/inactive_channels clarify which balance fields each count contributes to.
- connected_peers renamed to connected_channel_peers to reflect that it only lists peers we have a channel with; dedup via HashSet.
- fees_collected_msat removed from NodeState (still available on GetInfoResponse); it is a routing-node concern that doesn't belong in the wallet balance snapshot.

Adds ChannelState::is_open and channel_payout_still_pending helpers, NAPI mirror types synced, and refreshes Python + Kotlin NodeState tests to match the new type shape.

### DIFF
--- a/libs/gl-sdk-android/lib/src/androidInstrumentedTest/kotlin/com/blockstream/glsdk/NodeOperationsTest.kt
+++ b/libs/gl-sdk-android/lib/src/androidInstrumentedTest/kotlin/com/blockstream/glsdk/NodeOperationsTest.kt
@@ -2,6 +2,8 @@ package com.blockstream.glsdk
 
 import android.system.Os
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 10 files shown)


cdecker:
The new channel state `Unknown` is not covered in a number of `match` statements, otherwise this is good to go 👍

angelix:
> The new channel state `Unknown` is not covered in a number of `match` statements, otherwise this is good to go 👍

Should be fixed.

cdecker:
aCK https://github.com/Blockstream/greenlight/pull/704/commits/d8823000920fc27b1ab91a3cb76e291fd0d2a5f0

PR #708, opened 2026-04-27
gl-sdk: Add LNURL-auth (LUD-04 / LUD-05) and unify Node entry points

angelix:
## Summary

- **LNURL-auth (LUD-04 / LUD-05).** New `Node::lnurl_auth(request)` signs the service's `k1` challenge and posts the LUD-04 callback. Per-domain LUD-05 derivation happens internally; callers don't supply any keys.
- **Hardened-path security model.** The `m/138'` xpriv is derived once from the BIP39 seed at register/recover/connect time and stored on `Node` in `Zeroizing<Vec<u8>>`. The seed itself never persists. `m/138'` is hardened, so the stored material cannot derive any other wallet key — blast radius is restricted to LNURL-auth identities. `disconnect()` and `Drop for Node` both `take()` the option to scrub eagerly.
- **Mainstream wallet HMAC convention.** LUD-05 uses the 32-byte private key at `m/138'/0` as the HMAC key (Phoenix / Mutiny / Zeus / BlueWallet form), enabling cross-wallet identity portability at LNURL-auth services. Locked in by a known-vector test.
- **`InputType::LnUrlAuth { data }`.** `parse_input` classifies `tag=login` URLs offline (no HTTP fetch). New types `LnUrlAuthRequestData` and `LnUrlCallbackStatus`.
- **Zero public Node constructors.** Removed the bare `Node(credentials)` from both UniFFI and napi surfaces. The crate-private constructor is renamed `with_signer` → `new` (`pub(crate)`). All bindings expose the same four free functions: `register / recover / connect / register_or_recover`.
- **napi gains parity.** Free fns `register / recover / connect / registerOrRecover`, a `Config` wrapper (`withNetwork`, `withDeveloperCert`), and `Node.disconnect()` / `Node.credentials()`. Tests and README rewritten — JS callers no longer manage an external Signer/Handle.

## Test plan

- [x] `cargo test -p gl-sdk --lib` — 28/28 pass (offline classification, LUD-05 derivation determinism, signature verifies against derived xpub, callback-status JSON parsing, fixed pubkey vector for the BIP39 abandon-mnemonic at `example.com`)
- [x] `cargo build -p gl-sdk` and `cargo build -p gl-sdk-node` — both clean
- [x] Python bindings regenerated; `glsdk.Node()` raises `ValueError` (no constructor); `register/recover/connect/register_or_recover/parse_input/lnurl_auth` exposed
- [x] TypeScript bindings regenerated; `register/recover/connect/registerOrRecover/parseInput` async free fns and `Node.lnurlAuth(request)` exposed
- [ ] Python integration tests (`tests/test_lnurl.py`, `tests/test_parse_input.py`) — locally blocked by Python 3.14 + coincurve wheel issue unrelated to this branch; should run in CI. New auth coverage: classification-without-HTTP, end-to-end success, deterministic per-domain pubkey, post-disconnect scrub. The two previously-skipped LNURL-pay end-to-end tests are un-skipped via a `_bip39_seed` helper that aligns the gl-client `clients` fixture with `glsdk.connect`'s mnemonic-derived seed
- [ ] Android instrumented tests (`LnurlParseTest.kt`) — 5 new offline LnUrlAuth parse cases; needs on-device or CI run
- [ ] napi jest tests — rewritten on top of the new register/recover/connect surface; needs CI (depends on `gltestserver`)


cdecker (changes requested):
No, the signer is not guaranteed to be at the same location. If we want to have lnurl-auth we will need to run it through the CLN node, where we may have a signer present.

The reason for wanting signerless clients is because the signer is equivalent to root access to the node, i.e., the signer can self-certify any runes and TLS certs to escalate whatever privileges. We cannot assume the signer is present, and any operation that requires signing something, must involve the node.

libs/gl-sdk/src/node.rs
```diff
@@ -693,12 +722,18 @@ impl Node {
         Ok(())
     }
 
-    /// Internal constructor used by the high-level register/recover/connect functions.
-    /// Creates a Node with credentials and signer handle attached.
-    pub(crate) fn with_signer(
+    /// Canonical constructor.
+    ///
+    /// Crate-private — UniFFI consumers reach this via the top-level
+    /// `register` / `recover` / `connect` / `register_or_recover` free
```
cdecker, line 1010: Notice that this runs into the same issue of signerless clients: `register`, `recover` and `register_or_recover` will require a signer, which may not always be present. So using them as a entrypoints to get a Node handle is not a good idea.

cdecker: And I just noticed that I overlooked that `connect` also is requiring the mnemonic, so this API just threw the signerless mode out of the window, which means everybody is running around with root privileges, for no apparent reason, other than simplicity.

We can keep it this way, but we'll want to allow signerless clients at some point, so we will need to bring back in.

cdecker (changes requested):

libs/gl-sdk/src/node.rs
angelix: Fixed in #709

cdecker:
Rebased on top of main, but deferring as it requires a signer at the moment, and signerless clients will not work with this.

PR #709, opened 2026-04-27
feat(gl-sdk): builder-style Node creation, signerless by default

angelix:
NodeBuilder is the sole public entry point for Node construction
across all foreign bindings. The former free functions register /
recover / connect / register_or_recover are demoted to crate-private
helpers (`*_internal`); foreign-binding consumers go through
`NodeBuilder` exclusively.

Two design rules drive the shape:

1. Naked free functions are hard to extend without semver breaks. A
   builder absorbs new modifiers as additional with_* setters —
   additive forever.
2. Signer access ≡ root access on the node (self-certifies runes,
   mints TLS certs). The SDK must support keyless clients that don't
   hold the seed at all (paired devices, browser extensions, hardware
   signers). Signerless connect must be a first-class path, not an
   afterthought.

Resulting surface:

    // Signerless connect — caller has no mnemonic in this process.
    // SDK runs no signer; signing happens at the CLN node, a paired
    // device, or hardware. The keyless-client model.
    let node = NodeBuilder::new(&config).connect(credentials, None)?;

    // Signed connect — caller hands the mnemonic per-call, SDK
    // spawns a signer for the lifetime of the Node.
    let node = NodeBuilder::new(&config).connect(credentials, Some(mnemonic))?;

    // Register / recover / register_or_recover require a mnemonic
    // by definition (the signer must sign the registration /
    // recovery challenge). Mnemonic is positional, not stateful.
    let node = NodeBuilder::new(&config)
        .with_event_listener(listener)
        .register(mnemonic, invite_code)?;

The mnemonic is never stored on the builder. It is a positional
argument on the build call that needs it, so its lifetime is bounded
to that call and there is no half-set state. Modifiers like
with_event_listener live on the builder; secrets do not.

Surface
- NodeBuilder::new(config) — collects config + optional modifiers,
  no I/O.
- with_event_listener(listener) — fluent setter, returns the same
  Arc<Self> so calls chain in every binding.
- register(mnemonic, invite_code) — mnemonic required.
- recover(mnemonic) — mnemonic required.
- register_or_recover(mnemonic, invite_code) — mnemonic required.
- connect(credentials, mnemonic: Option<String>) — mnemonic optional;
  None produces a signerless Node.

Implementation notes
- Node::signerless(credentials) is a new pub (non-UniFFI-export)
  Rust constructor. Used by the builder for the None-mnemonic path
  and by gl-sdk-napi to back its `new Node(credentials)` constructor
  with the same signerless semantics.
- crate::connect_signerless_internal wires Node::signerless into
  the lib.rs internals; crate::connect_internal continues to drive
  the signed connect via the SDK signer spawn.
- Listener installation goes through Node::set_event_listener
  (pub(crate)) which spawns a background tokio task that tails the
  gRPC event stream and dispatches to listener.on_event. The task
  is aborted on Drop and replaced if a new listener is set.
- The polling-style Node::stream_node_events() API stays for callers
  who prefer to drive events themselves; the builder route is just
  a callback-style alternative that can't miss early events.
- Listeners are taken out of the builder state on each build call,
  so reusing a NodeBuilder for a second build requires re-setting
  the listener.

Tests migrated
- Python: test_auth_api.py (24 callsites), test_list_payments.py
  (6), test_node_methods.py (2).
- Kotlin: AuthApiTest.kt (8), NodeOperationsTest.kt (2),
  ListPaymentTest.kt (3), LoggingTest.kt (1). All converted to the
  positional-mnemonic-on-build-call shape.

Verified: `cargo build -p gl-sdk -p gl-sdk-node` clean; Python
bindings expose `NodeBuilder` (with the new method signatures) and
`NodeEventListener` and no longer expose the demoted free
functions.

### DIFF
--- a/libs/gl-sdk/src/node_builder.rs
+++ b/libs/gl-sdk/src/node_builder.rs
@@ -1,0 +1,162 @@
+// Builder-style Node creation.
+//
+// `NodeBuilder` is the sole public entry point for Node construction
+// across all foreign bindings. The shape is "request the action,
+// optionally with modifiers":
+//
+//     // Signerless connect — caller does not have the mnemonic in
+//     // this process. The SDK runs no signer; signing happens
+//     // elsewhere (paired device, CLN node's local signer, hardware
+//     // signer). This is the supported model for keyless clients.
+//     let node = NodeBuilder::new(&config).connect(credentials, None)?;
+//
+//     // Signed connect — caller hands the mnemonic per build call,
+//     // SDK spawns a signer.
+//     let node = NodeBuilder::new(&config).connect(credentials, Some(mnemonic))?;
+//
+//     // Register / recover require a mnemonic by definition (the
+//     // signer must sign the registration/recovery challenge).
+//     let node = NodeBuilder::new(&config)
+//         .with_event_listener(listener)
+//         .register(mnemonic, invite_code)?;
+//
+// The mnemonic is the security-sensitive input: hand it in only when
+// you actually want the SDK to act as a signer for that call.
+//
+// Adding a new modifier later is a new `with_*` setter — additive,
+// never breaks existing callers.
+
+use std::sync::Arc;
+
+use crate::{
+    config::Config,
+    node::{Node, NodeEventListener},
+    Error,
+};
+
+/// Configurable Node construction. See module docs.
+///
+/// All fields are immutable after construction. Each `with_*` setter
+/// returns a fresh `Arc<NodeBuilder>` that shares ownership of any
+/// previously-installed modifiers via `Arc<dyn …>`. No interior
+/// mutability, no locks — the builder is a value, not a state
+/// machine.
+#[derive(uniffi::Object)]
+pub struct NodeBuilder {
+    config: Arc<Config>,
+    event_listener: Option<Arc<dyn NodeEventListener>>,
+}
+
+#[uniffi::export]
+impl NodeBuilder {
+    /// Create a builder for a Node with `config`. No I/O happens
+    /// until you call `connect` / `register` / `recover` /
+    /// `register_or_recover`.
+    #[uniffi::constructor]
+    pub fn new(config: &Config) -> Arc<Self> {
+        Arc::new(Self {
+            config: Arc::new(config.clone()),
+            event_listener: None,
+        })
+    }
+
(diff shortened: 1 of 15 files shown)


libs/gl-sdk/src/node_builder.rs
```diff
@@ -0,0 +1,136 @@
...
+    event_listener: Option<Box<dyn NodeEventListener>>,
+}
+
+#[uniffi::export]
+impl NodeBuilder {
+    /// Create a builder for a Node tied to `mnemonic` and `config`.
+    /// No I/O happens until you call `register` / `recover` /
+    /// `connect` / `register_or_recover`.
+    #[uniffi::constructor]
+    pub fn new(mnemonic: String, config: &Config) -> Arc<Self> {
```
cdecker: Yeah, this makes the seed / mnemonic mandatory, which means we no longer have signerless clients.

libs/gl-sdk/src/node_builder.rs
```diff
@@ -0,0 +1,136 @@
...
+/// `register_or_recover` free functions don't fit — typically because
+/// you need to install per-node observers (currently: an event
+/// listener) before the underlying gRPC stream is established, so
+/// early events can't be missed.
+///
+/// New modifiers are exposed as additional `with_*` setters; existing
+/// callers stay source-compatible.
+#[derive(uniffi::Object)]
+pub struct NodeBuilder {
+    state: Mutex<BuilderState>,
```
cdecker: Not quite sure why this needs to be wrapped in a mutex?

cdecker: Is it because it is a `uniffi::Object` which implies `Clone` maybe?

cdecker (commented):
Just a minor thing about losing the Signerless support, otherwise good to go :+1:

libs/gl-sdk/src/node_builder.rs
angelix: Added support to signerless clients

libs/gl-sdk/src/node_builder.rs
angelix: Changed the implementation in f202cb3c5e3a24dea8ab5414e46b600f8bcc7ff4

cdecker:
Lovely, thanks @angelix :hugs:

PR #716, opened 2026-05-13
sdk: Stable numeric error codes for i18n support

cdecker:
## Summary

- Replace gl-sdk tuple-style error variants with named-field variants carrying `code: i32`, `message: String`, and `values: HashMap<String, String>`
- Each error has a stable numeric code (1000-1099 node identity, 1100-1199 credentials, 2000-2099 RPC, 3000-3099 validation, 9000-9099 generic) enabling host-language i18n lookup
- The `values` map provides structured interpolation keys so callers can fill translated templates (e.g., `"Node {node_id} already exists"`)
- Constructor functions (`Error::rpc()`, `Error::other()`, etc.) keep call sites clean
- Regenerated Python, Kotlin, Swift, and Ruby bindings — all expose `.code`, `.message`, `.values` on error subclasses

## Test plan

- [x] `cargo build -p gl-sdk` compiles cleanly
- [x] Verified generated Python bindings have `code`, `message`, `values` attributes on all `Error` subclasses
- [x] Verified generated Kotlin bindings have `val code: Int`, `val message: String`, `val values: Map<String, String>`
- [x] Verified generated Swift bindings have `code: Int32, message: String, values: [String: String]`
- [ ] Integration test with host-language error catching and i18n template interpolation

🤖 Generated with [Claude Code](https://claude.com/claude-code)


michaelbship:
@cdecker we will follow this up with a doc fix as well

cdecker:
Rebased on top of `main`. Changes during rebase:

- Fixed all callers added on `main` since the original PR (`credentials.rs`, `input.rs`, `lib.rs`, `logging.rs`, `node.rs`) — they used old tuple-style `Error::Other(...)` / `Error::Rpc(...)` etc. and now use the new constructor methods (`Error::other()`, `Error::rpc()`, `Error::duplicate_node()`, etc.)
- Removed the stray `bindings/glsdk.py` — Python bindings belong in `glsdk/`, other languages in `bindings/`
- Moved all derived files (`glsdk.py` + Kotlin/Swift/Ruby) into the second commit, keeping commit 1 purely Rust source changes
- Regenerated `glsdk.py` from the updated Rust definitions

PR #717, opened 2026-05-15
Command hex format

Lagrang3:
Added json outputs to `gl-cli` commands.

For example:
```
$ glcli --json node getinfo
{
  "address": [],
  "alias": "...",
  "binding": [],
  "blockheight": 949533,
  "color": "...",
  "fees_collected_msat": 0,
  "id": "...",
  "lightning_dir": "/tmp/bitcoin",
  "network": "bitcoin",
  "num_active_channels": 0,
  "num_inactive_channels": 0,
  "num_peers": 1,
  "num_pending_channels": 0,
  "our_features": {
    "channel": "",
    "init": "0898880a8a59a1",
    "invoice": "02000002024100",
    "node": "8898880a8a59a1"
  },
  "version": "v25.12gl1"
}
```

I did this by hand, for every possible gRPC response that we use, which are not all of CLN's commands.
I know this is flawed, since we have no guarantee of correctedness by developer error or API changes and
as the number of gRPC commands grow this task becomes tedious.
But right now it is manageable and until we find a trick to make it automatic we can use this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1522,6 +1522,7 @@ dependencies = [
  "futures",
  "gl-client",
  "hex",
+ "serde_json",
  "thiserror 2.0.18",
  "tokio",
  "vls-core",
--- a/libs/gl-cli/Cargo.toml
+++ b/libs/gl-cli/Cargo.toml
@@ -28,6 +28,7 @@ hex = "0.4"
 thiserror = "2.0.11"
 tokio = "1.43.0"
 vls-core.workspace = true
+serde_json = "^1.0"
 
 [badges]
 maintenance = { status = "actively-developed" }
--- a/libs/gl-cli/src/bin/glcli.rs
+++ b/libs/gl-cli/src/bin/glcli.rs
@@ -1,14 +1,21 @@
 use clap::Parser;
 use gl_cli::{run, Cli};
+use serde_json::json;
 
 #[tokio::main]
 async fn main() {
     let cli = Cli::parse();
+    let json_print = cli.json;
 
     match run(cli).await {
         Ok(()) => (),
         Err(e) => {
-            println!("{}", e);
+            if json_print {
+                let j = json!({"error": e.to_string()});
+                println!("{}", serde_json::to_string_pretty(&j).unwrap());
+            } else {
+                println!("{}", e);
+            }
         }
     }
 }
--- a/libs/gl-cli/src/error.rs
+++ b/libs/gl-cli/src/error.rs
@@ -16,6 +16,9 @@ pub enum Error {
 
     #[error(transparent)]
     UtilError(#[from] util::UtilsError),
+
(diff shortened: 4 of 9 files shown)


cdecker:
Very good idea, the JSON output is likely more useful than the native protobuf serialization. However, I know that there are ways to add attributes to the structs that are generated in Rust. The App team had used it a while ago to convert arbitrary messages to JSON for their own logging.

They were using the `message_attribute()` builder call to have the generated messages automatically be annotated with `#[derive(Serialize, Deserialize)]` and then did something to convert binary fields to `hex`, I think with a custom serializer that knew how to identify and handle the binary values. This would solve the manual maintenance you mention above and absolutely ensure that the JSON and binary representation remain in sync.

cdecker:
Let me know if you'd like to merge this as is, so the functionality is available, or whether you want to look into a more complex, but simpler longer term, solution :-)

Lagrang3:
For me, this is the best we have so far.
So I am good with merging now and then in a follow up PR try to make it more general.

Lagrang3:
> Very good idea, the JSON output is likely more useful than the native protobuf serialization. However, I know that there are ways to add attributes to the structs that are generated in Rust. The App team had used it a while ago to convert arbitrary messages to JSON for their own logging.
> 
> They were using the `message_attribute()` builder call to have the generated messages automatically be annotated with `#[derive(Serialize, Deserialize)]` and then did something to convert binary fields to `hex`, I think with a custom serializer that knew how to identify and handle the binary values. This would solve the manual maintenance you mention above and absolutely ensure that the JSON and binary representation remain in sync.

I have to look more deeply into this.
AFAIK the custom serializer implies redoing everything by hand again, unless we write some program that writes code.
The rust pb structs already convert to json with `serde_json::to_string(response)` for example, but with the
annoying byte arrays represented as arrays of integers instead of hex, `amount_msat` represented as an object with an
`msat` field instead of an integer and None fields are also printed with the value `none`.

UPD: I also noticed that the default serialization leads to enum types to appear in their integer form, eg. `"status": 2`, instead of `"status": "complete"`.

cdecker:
Indeed, that was the same problem I had when we first built cln-grpc, which required converting to grpc (which we wanted to do) and then back from grpc to JSON (mostly for testing so we could run tests via direct UDS JSON-RPC and grpc unchanged), but the errors multiply aloong the conversion chain, making this rather difficult.

I wonder if it would make sense to also have the conversion back to JSON-RPC in cln-grpc, so we could then do roundtrip testing.

FWIW implementing a serializer is not that much work, I think the central dispatch `serialize` method is the one we'd have to add a match statement and then handing on to the existing serialization methods (doing everything like the JSON serializer, but detecting binary and hex-encoding those first). Doesn't work all that well for enums as you correctly pointed out :thinking: 

Agreed, let's merge as is and start using it, and then we can revisit if the maintenance burden gets too much.

PR #718, opened 2026-05-18
gl-sdk: Add C++ bindings

ardier16:
Adds C++ bindings support to `gl-sdk`:

- Introduce a `cpp-bindings` feature and C++ compatible `resolve_input` export.
- Add `task sdk:bindings-cpp` and include it in `task sdk:bindings-all`.
- Use `uniffi-bindgen-cpp` plus `perl` post-processing to rename generated `register` identifiers to `register_node`.
- Update the SDK README with C++ bindings setup and generation instructions.

### DIFF
--- a/libs/gl-sdk/src/lib.rs
+++ b/libs/gl-sdk/src/lib.rs
@@ -259,9 +259,18 @@ pub fn parse_input(input: String) -> Result<input::ParsedInput, Error> {
 /// immediately without I/O.
 ///
 /// Strips `lightning:` / `LIGHTNING:` prefixes automatically.
+#[cfg(not(feature = "cpp-bindings"))]
 #[uniffi::export(async_runtime = "tokio")]
 pub async fn resolve_input(input: String) -> Result<input::ResolvedInput, Error> {
     input::resolve_input(input).await
+}
+
+/// Synchronously classify and resolve the input (C++ bindings).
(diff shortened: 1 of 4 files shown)


cdecker:
Excellent addition @ardier16, thank you :pray: 

Can we add a generation run in CI as a followup, so we can ensure that the CPP stub generation works as expected, and we don't accidentally break it?

libs/gl-sdk/src/lib.rs
```diff
@@ -259,11 +259,20 @@ pub fn parse_input(input: String) -> Result<input::ParsedInput, Error> {
 /// immediately without I/O.
 ///
 /// Strips `lightning:` / `LIGHTNING:` prefixes automatically.
+#[cfg(not(feature = "cpp-bindings"))]
```
cdecker, line 262: We can make the entirety of the API sync, and then running async code inside of the Rust crate only. This would mean all the APIs are the same regardless of the host language are sync-only.

cdecker (approved):

libs/gl-sdk/src/lib.rs
cdecker: I'll do this in a followup PR. As @ardier16 pointed out in a call this is a breaking change, and together with #716 we will bump version to 0.4 once this is merged.

PR #729, opened 2026-06-09
chore: add v26.06gl1 version

Lagrang3:


Lagrang3:
Rebased

cdecker (approved):
Excellent work, thanks for removing that deprecated call, and I like that this is our first multi-version client too :-)

PR #742, opened 2026-08-25
gl-plugin: make lsp_invoice RPC call store additional invoice metadata

Nazarevsky:
# Problem
Currently, when a client requests for lsp invoice (JIT channel opening), no information about the original invoice is stored. Basically, the client hands out the requested invoice to a payer but receives only reduced amount. We'd like to have a way of understanding whether the requested invoice was for channel opening or a simple payment.

# Solution
Introduced a structure `LspInvoiceMeta` which contains all information that we need to store (specifically, bolt11 invoice). Use datastore requests in order to store metadata. `LspInvoiceMeta` is stored in both cases - if the invoice was requested from the LSP or not.

# Changes
- introduced `LspInvoiceMeta` structure for storing meta
- updated lsp_invoice RPC call to store `LspInvoiceMeta`

### DIFF
--- a/libs/gl-plugin/src/node/mod.rs
+++ b/libs/gl-plugin/src/node/mod.rs
@@ -1,6 +1,6 @@
 use crate::config::Config;
 use crate::pb::{self, node_server::Node};
-use crate::storage::{LspInvoiceMeta, StateStore};
+use crate::storage::{JitRequestMeta, StateStore};
 use crate::{messages, Event};
 use crate::{stager, tramp};
 use anyhow::{anyhow, Context, Error, Result};
@@ -350,7 +350,7 @@ impl Node for PluginNodeServer {
         // A JIT channel has now been negotiated with the LSP for this
         // invoice. So, we're storing some data with the original requested
         // amount.
-        let meta = LspInvoiceMeta {
+        let meta = JitRequestMeta {
             label: invoice_label.clone(),
             payment_hash: res.payment_hash.clone(),
             requested_amount_msat,
@@ -859,7 +859,7 @@ impl Node for PluginNodeServer {
 
 /// Writes `LspInvoiceMeta` using datastore request. `LspInvoiceMeta` is useful for defining
 /// some additional information regarding invoice being requested trough Greenlight.
-async fn write_lsp_invoice_meta(rpc: &mut ClnRpc, meta: LspInvoiceMeta)
+async fn write_lsp_invoice_meta(rpc: &mut ClnRpc, meta: JitRequestMeta)
                                 -> Result<()> {
     let record_serialized =
         serde_json::to_string(&meta).context("failed to serialize LspInvoiceMeta")?;
--- a/libs/gl-plugin/src/node/mod.rs
+++ b/libs/gl-plugin/src/node/mod.rs
@@ -270,11 +270,14 @@ impl Node for PluginNodeServer {
                     label: req.label.clone(),
                     payment_hash: res.payment_hash.to_string(),
                     requested_amount_msat: req.amount_msat,
+                    expected_amount_msat: 0,
                     bolt11: res.bolt11.clone(),
+                    lsp_id: "".to_string(),
                 };
 
-                write_lsp_invoice_meta(rpc, meta).await
-                    .map_err(|e| Status::new(Code::Internal, e.to_string()))?;
+                if let Err(e) = write_lsp_invoice_meta(&mut rpc, meta).await {
+                    warn!("Failed to write LSP invoice meta: {}", e);
+                }
 
                 return Ok(Response::new(pb::LspInvoiceResponse {
                     bolt11: res.bolt11,
@@ -364,11 +367,14 @@ impl Node for PluginNodeServer {
             label: invoice_label.clone(),
             payment_hash: res.payment_hash.clone(),
             requested_amount_msat,
+            expected_amount_msat: requested_amount_msat.saturating_sub(opening_fee_msat),
             bolt11: res.bolt11.clone(),
+            lsp_id,
         };
 
-        write_lsp_invoice_meta(rpc, meta).await
-            .map_err(|e| Status::new(Code::Internal, e.to_string()))?;
+        if let Err(e) = write_lsp_invoice_meta(&mut rpc, meta).await {
+            warn!("Failed to write LSP invoice meta: {}", e);
+        }
 
         Ok(Response::new(res.into()))
     }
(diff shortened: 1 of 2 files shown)


libs/gl-plugin/src/node/mod.rs
```diff
@@ -266,6 +266,16 @@ impl Node for PluginNodeServer {
...
+                let meta = LspInvoiceMeta {
+                    label: req.label.clone(),
+                    payment_hash: res.payment_hash.to_string(),
+                    requested_amount_msat: req.amount_msat,
+                    bolt11: res.bolt11.clone(),
+                };
+
+                write_lsp_invoice_meta(rpc, meta).await
+                    .map_err(|e| Status::new(Code::Internal, e.to_string()))?;
+
```
nepet: I don't think we need to store the invoice meta data for a regular invoice. It will just double what we already have. Storing it for a jit-channel invoice is enough to identify it as such: If it's in the datastore, it automatically is a jit-channel invoice.

libs/gl-plugin/src/storage.rs
```diff
@@ -66,3 +67,22 @@ impl StateStore for SledStateStore {
...
+/// structure instead of querying invoices table is that we want to distinguish
+/// incomming payments whether they were for JIT channel opening or just a simple
+/// payment. Currently, CLN does not allow to do so, that's why this workaround
+/// exists.
+#[derive(Serialize, Deserialize)]
+pub struct LspInvoiceMeta {
+    pub label: String,
+    pub payment_hash: String,
+    pub requested_amount_msat: u64,
+    pub bolt11: String,
```
nepet, line 94: nit: I think we could also store the `peer_id` and the `expected_amount_msat`. But feel free to leave it out if it's not required.

libs/gl-plugin/src/node/mod.rs
```diff
@@ -343,6 +356,20 @@ impl Node for PluginNodeServer {
...
+        // amount.
+        let meta = LspInvoiceMeta {
+            label: invoice_label.clone(),
+            payment_hash: res.payment_hash.clone(),
+            requested_amount_msat,
+            bolt11: res.bolt11.clone(),
+        };
+
+        write_lsp_invoice_meta(rpc, meta).await
+            .map_err(|e| Status::new(Code::Internal, e.to_string()))?;
```
nepet: Do we really want to fail the whole RPC call in the (unexpected) case that we can't write the meta data? We already negotiated a jit-channel invoice and stored it on CLN. I think it's enough to just `log::warn` and continue. We won't loose much.

libs/gl-plugin/src/node/mod.rs
```diff
@@ -837,6 +864,33 @@ impl Node for PluginNodeServer {
     }
 }
 
+/// Writes `LspInvoiceMeta` using datastore request. `LspInvoiceMeta` is useful for defining
+/// some additional information regarding invoice being requested trough Greenlight.
+async fn write_lsp_invoice_meta(mut rpc: tokio::sync::MutexGuard<'_, ClnRpc>, meta: LspInvoiceMeta)
```
nepet: nit: The signature is a bit weird, compared to the other callers of `ClnRpc` in this file. This function takes ownership of the lock, which means that the caller can't use `rpc` afterwards. Every other helper in this file takes `&mut rpc`:
```
async write_lsp_invoice_meta(rpc: &mut ClnRpc, meta: LspInvoiceMeta)
```

libs/gl-plugin/src/node/mod.rs
```diff
@@ -837,6 +864,33 @@ impl Node for PluginNodeServer {
...
+                                -> Result<()> {
+    let record_serialized = serde_json::to_string(&meta)
+        .map_err(|e| Status::new(Code::Internal, e.to_string()))?;
```
nepet: nit: This function returns an `anyhow::Result`. The extra serialization into a `tonic::Status` seems a bit useless. It's coerced into `anyhow::Error` and the caller converts it back into a `tonic::Status`.

```suggestion
    let record_serialized =
        serde_json::to_string(&meta).context("failed to serialize LspInvoiceMeta")?;
```

libs/gl-plugin/src/node/mod.rs
```diff
@@ -837,6 +864,33 @@ impl Node for PluginNodeServer {
...
+
+    let datastore_req = cln_rpc::model::requests::DatastoreRequest {
+        key: vec![
+            "gl".to_string(),
+            "jit_channels".to_string(),
+            meta.label,
+        ],
```
nepet, line 872: Depending on what info the reading part actually has, I'd rather use the `payment_hash` as the preferred key instead of the `label` -  which by the way can be an ugly (maybe even unbounded) string.

```suggestion
    let datastore_req = cln_rpc::model::requests::DatastoreRequest {
        key: vec![
            "gl".to_string(),
            "lsp_invoices".to_string(),
            meta.payment_hash,
        ],
```

nepet (changes requested):
Thank you @Nazarevsky for putting this together.

The approach makes sense and the doc comment on `LspInvoiceMeta` does a good job explaining *why* this workaround is needed, which will save the next person a lot of time.

One general note to keep in mind, not a blocker: nothing cleans these entries up on invoice expiry or payment so far, so they accumulate indefinitely. May not be a problem though, if we only store "lsp_invoices" and if we don't expect them to be created indefinitely.

libs/gl-plugin/src/node/mod.rs
Nazarevsky: Initially I thought of using payment hashes here but was aware of some ambiguity between invoices.

Since the original and reduced invoices have different amounts, payment hashes will also be different. In other words, we're storing an invoice with the key containing the original invoice. On GL client side when we intercept the invoice payment, the payment hash corresponds to the reduced invoice. Therefore, we'll fail to retrieve the original invoice by the reduced payment hash.

My intention about using invoice label was in it's nature - it is unique among all created invoices. That means, both original and reduced invoice should have the same label (uniqueness applies between original and reduced invoices because the original invoice invoice is not stored on client). It's indeed an ugly way, but I'm not sure about any other ways

libs/gl-plugin/src/storage.rs
Nazarevsky: I think it make sense to add `lsp_id` and `expected_amount_msat`. Even since `lsp_id` can be obtained from the bolt11 invoice, it will be much harder to do it with `expected_amount_msat`.

cdecker:
@nepet I think this looks quite good now, want me to merge it?

nepet (approved):
Thanks @Nazarevsky for this important addition to create a more detailed view of the LSP interactions. LGTM

